PR #19, opened 2015-05-08
Update the example as it used archaic names

mtl2034:
Use parseGetPushSocket and parseProcessNextPushNotification in the example code. Original names never shipped.


hramos:
Looks like this is a good fix, but the PR was opened prior to the git commit history getting purged. Any chance you can open this as a new PR? I apologize for the inconvenience.

PR #26, opened 2015-05-13
Restore the Data guide

hramos:
Includes more information on data types, imports, and exports. Previously present at parse.com/docs/data

### DIFF
--- a/en/common/data.mdown
+++ b/en/common/data.mdown
@@ -1,0 +1,109 @@
+# Data
+
+## Valid Data Types
+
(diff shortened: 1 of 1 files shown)


hramos:
I'm guessing this will require an update to Parse Config to include the new Data section at the end.

Mattieuga:
Cool, you'll just have to add "data" to the config toc. I can run through that with you tomorrow.

hramos:
"Data" has been added to the Parse Config TOC.

PR #30, opened 2015-05-14
Replace all usages of PFUser with token in iOS guides.

nlutsenko:


hramos:
What text will be shown when this token is rendered?

nlutsenko:
Merging this PR into PR #31

PR #31, opened 2015-05-14
Replace all usages of PFInstallation/PFObject with token in iOS guides.

nlutsenko:


nlutsenko:
> What text will be shown when this token is rendered?
Answering here, sorry...
It will be replaced with `PFUser` as anything that comes like %{TOKEN} is going to be replaced according to the map in `tokens.json`.

PR #32, opened 2015-05-14
Fixed whitespace to avoid weird line breaks in code.

nlutsenko:


nlutsenko:
This gets rendered in a weird way on https://parse.com/docs/ios/guide
It looks like it tries to add the code block to `<li>` tag which causes the line breaks to behave in a weird way. Not 100% sure if it will fix it, but I sincerely hope so.

PR #38, opened 2015-05-15
Fix setRoleWriteAccess  typo in "Security for Other Objects" documentation

daviddoran:
Bug reported here: https://developers.facebook.com/bugs/1443247899303928/

To set role access by name we need to use `setRoleWriteAccessWithName`.


gfosco:
Thanks!

PR #43, opened 2015-05-16
Docs - incorrect function definition

craiggrummitt:
parseClassName returns String, not String!


hramos:
@craiggrummitt can you email me at hector@parse.com with your t-shirt size and shipping address? We'd like to send you some swag.

PR #45, opened 2015-05-18
Add swift examples

wiruzx:


hramos:
@wiruzx can you email me at hector@parse.com with your t-shirt size and shipping address? We'd like to send you some swag.

Issue #21, opened 2015-05-12
Swift sample code missing for various sections


hramos:
Some of the files in the iOS guide are missing Swift examples (search for "no swift example"):

Docs/en/ios/handling-errors.mdown
Docs/en/ios/in-app-purchases.mdown
Docs/en/ios/local-datastore.mdown
Docs/en/ios/push-notifications.mdown
Docs/en/ios/queries.mdown
Docs/en/ios/user-interface.mdown

Lucashuang0802:
Finished editing from the item one to item four in pull request #41, please review and approve if that's what you guys need. And I will finish the rest.

hramos:
Linking #45 and #47 here.

Lucashuang0802:
Link #58

Lucashuang0802:
#73 This issue can be close

nlutsenko:
It looks like there are couple more missing:
https://github.com/ParsePlatform/Docs/blob/a9ab036799839dd94badad00341f7f5ccba041f2/en/ios/push-notifications.mdown#L685
https://github.com/ParsePlatform/Docs/blob/2aa74b0d90d48edc662193085a8fda8c7bae5df2/en/common/relations.mdown#L201
https://github.com/ParsePlatform/Docs/blob/2aa74b0d90d48edc662193085a8fda8c7bae5df2/en/common/relations.mdown#L249

Lucashuang0802:
I will look into it when I have time tonight and hopefully I can make PR tomorrow. Thanks.

hramos:
Thanks for all your help, it's really appreciated.

Lucashuang0802:
#78 is merged. Please check whether there are still missing pieces of **Swift** codes for **iOS**

hramos:
Awesome work! This is really appreciated.

I think the only missing pieces are in `en/common/relations.mdown`, which I missed in my initial pass when this issue was opened.

Lucashuang0802:
Okay, I will take a look at that.

Lucashuang0802:
I just made a double check and I can't find any `// no swift example` in the `en/common/relations.mdown`. Can you double check? Thanks.

hramos:
Looks like we now have full Swift coverage in the Docs. Thank you very much!

Lucashuang0802:
Sounds good. @ me if you any of them. lol. Thanks.

PR #49, opened 2015-05-19
Update files.mdown

xpereta:
Repaired the saveInBackgroundWithBlock example to work with Xcode 6.3.1


xpereta:
I have noticed that this part of the examples does not work with the current version of Xcode (6.3.1) It would be better to review all docs to update them to the version that the users are most likely using.

hramos:
Linking to #21

nlutsenko:
Looks great! Thanks!

hramos:
Deployed.

hramos:
@xpereta can you email me at hector@parse.com with your t-shirt size and shipping address? We'd like to send you some swag.

PR #55, opened 2015-05-21
Actualization documentation

Fi5t:
Update of russian version of documentation based on english version.


hramos:
Much appreciated!

PR #57, opened 2015-05-21
Users - function name/definition corrections

craiggrummitt:


craiggrummitt:
Sure, updated with trailing closure syntax. (Agree this approach is 'better' practice, but I hadn't done it this way for consistency with the rest of the docs)

nlutsenko:
We'll get there :D Thanks for updating. Merging it in now.

PR #61, opened 2015-05-26
Change incorrect string concatenation character

raidenace:
In the very first PHP code example, the string concat character is given as "+" in the code. It should be "." I think this is critical, because most users might copy-paste the first example and might get "0" if there was an exception, instead of the actual message.


nlutsenko:
Awesome! Thanks!

PR #63, opened 2015-05-27
Add docs for the Schema API

sheki:


en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
+# Schema
+
+This API allows you to access the Schema of your app.
```
jkarraker: I think it should be schema rather than Schema. Lowercase is better cause it's not a proper noun.

Also we need to decide between schema and schemas. I think a single schema is for a single class, and so on this line it would be schemas.

en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
...
+          "type": "Date"
+        }
+      }
+    },
+    ...
+  ]
+}
+```
+
+To fetch schema of a particular class run
```
jkarraker: All sentences preceding a code block should end with :

en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
...
+       "X-Parse-REST-API-Key": "${REST_API_KEY}",
+       "Content-Type": "application/json"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+```
+
+## Adding a class
+
+To add a new class to your schema run.
```
jkarraker: Needs : 

This sentence is confusing because earlier you said fetch a schema, now you say add a class to your schema. So schema and class are referring to the same thing. I think here you need to say 'add a new schema'. But this sentence could use some more explanation of what this actually does.

en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
...
+       "X-Parse-REST-API-Key": "${REST_API_KEY}",
+       "Content-Type": "application/json"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+```
+
+## Modifying the schema
+
+To add fields to an existing class you can run
```
jkarraker: Again, there's confusing between schema/class, which you are using interchangeably.

en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
...
+     }), {
+       "X-Parse-Application-Id": "${APPLICATION_ID}",
+       "X-Parse-REST-API-Key": "${REST_API_KEY}",
+       "Content-Type": "application/json"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+```
+
+To delete a particular field you need to use `{"__op" : "Delete" }`
```
jkarraker: A little more explanation

en/rest/schemas.mdown
```diff
@@ -0,0 +1,186 @@
...
+       "X-Parse-REST-API-Key": "${REST_API_KEY}",
+       "Content-Type": "application/json"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+```
+
+## Removing a class
+
+You can only remove a class from your schema if it is empty (has 0 objects).
```
jkarraker: Another confusing schema/class mixup.

Explain how to empty a class.

lacker:
"You can only remove a schema room your app" - room?

lacker:
You can do a multi-line -d to make it easier to read. I suggest doing that for any examples that would otherwise require right-scrolling.

hramos:
I'll take a pass soon. The last step before merging this in would be to squash all the commits into one btw.

sheki:
Squashed commits into a single one.

en/rest/schemas.mdown
```diff
@@ -0,0 +1,196 @@
...
+custom views of your data. We use the schema API to display columns names and 
+types in the databrowser.
+
+This API allows you to access the schemas of your app.
+Note: This API can be only accessed using the `master key`.
+
+## Fetch the schema
+To fetch the Schema for all the classes of your app, run:
+
+Note: `createdAt` and `updatedAt` are of type. But they are not represented
```
hramos: "are of type" may be missing __Date at the end of the sentence?

en/rest/schemas.mdown
```diff
@@ -0,0 +1,196 @@
...
+types in the databrowser.
+
+This API allows you to access the schemas of your app.
+Note: This API can be only accessed using the `master key`.
+
+## Fetch the schema
+To fetch the Schema for all the classes of your app, run:
+
+Note: `createdAt` and `updatedAt` are of type. But they are not represented 
+as strings in objects. This is a special case for the parse API.
```
hramos: Capitalize Parse.

PR #65, opened 2015-05-27
this was a typo

aallen-dev:
As evidenced by the same sentence later in the parse website (where I read this document) the line should read "just by" and not "just be".


nlutsenko:
Good catch! Thanks!

PR #80, opened 2015-05-30
Clarify single user acl doesn't need logging in

kayalew:
Docs say that users must be logged in before creating single-user-acl.
API (https://parse.com/docs/js/api/symbols/Parse.ACL.html#constructor)
doesn't indicate that


kayalew:
Awesome!

PR #81, opened 2015-06-01
JS Using Relations

parties:


en/common/relations.mdown
```diff
@@ -445,8 +460,16 @@ let query = relation.query()
 // Want to contribute to this doc? https://github.com/ParsePlatform/Docs
 ```
 ```common-js
-// no js example
-// Want to contribute to this doc? https://github.com/ParsePlatform/Docs
+// suppose we have a book object
+var book = new Parse.Object('Book');
```
gfosco: I'd make this like the above example, assuming a previous object, since a new object will not have any relations.

    var book = ...

gfosco:
Closing due to lack of response.

PR #82, opened 2015-06-02
Add /1/apps API to REST docs

jkarraker:
API will launch in the next couple days

### DIFF
--- a/en/rest/apps.mdown
+++ b/en/rest/apps.mdown
@@ -8,10 +8,13 @@ the settings of an app that you are a collaborator on. This API can be used to
 create test apps to run in automated tests, and in combination with the schema
 API can be used to programmatically replicate your existing app into a test app.
 
-Authentication for the apps endpoint is done a little differently thn the rest
+Authentication for the apps endpoint is done a little differently than the rest
 of the REST API. Instead of authenticating with one of your app's keys, instead
 you must use your account's login info. The `X-Parse-Email` header identifies which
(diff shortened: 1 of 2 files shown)


en/rest/apps.mdown
```diff
@@ -0,0 +1,151 @@
...
+
+You can view, create, and edit your Parse apps via the REST API, in addition
+to through your parse.com dashboard. By authenticating with your Parse
+account's email and password, you can fetch all the app keys for each of the
+apps that you are a collaborator on, create a new app that you own, or update
+the settings of an app that you are a collaborator on. This API can be used to
+create test apps to run in automated tests, and in combination with the schema
+API can be used to programmatically replicate your existing app into a test app.
+
+Authentication for the apps endpoint is done a little differently thn the rest
```
hramos: s/thn/than/

en/rest/apps.mdown
```diff
@@ -0,0 +1,151 @@
...
+of the REST API. Instead of authenticating with one of your app's keys, instead
+you must use your account's login info. The `X-Parse-Email` header identifies which
+account you are using, and the `X-Parse-Password` header authenticates the endpoint.
```
hramos: Should we add a note saying that if your app has no username or password set, they should set one up at parse.com/account/edit ? This would be mainly for accounts that signed up with Facebook/GitHub/Google.

hramos:
LGTM. When do you want this to go live?

PR #87, opened 2015-06-10
Fix sample code for common performance section

chadedrupt:
Updates the js samples to be referenced as common-js.

And also includes sample code for Objective-C and Swift.

Will be able to add the other language samples soon.

### DIFF
--- a/en/common/performance.mdown
+++ b/en/common/performance.mdown
@@ -46,28 +46,37 @@ The order of a query constraint's usefulness is:
 
 Take a look at the following query to retrieve GameScore objects:
 
-```objc
+```common-js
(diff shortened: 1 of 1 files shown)


hramos:
This is great!

One nit: can you update the swift code block tags to be common-swift as well?

chadedrupt:
Done. not sure how I missed them 😄 Thanks

hramos:
Awesome, this all looks good to go. One last thing if you don't mind, can you squash these three commits into one?

You should be able to do so by running:

     $ git rebase -i HEAD~3

Since these commits are already associated with this pull request, you may need to use --force when pushing them:

     $ git push origin --force

chadedrupt:
There we go.

PR #89, opened 2015-06-10
pointer permissions docs

jkarraker:
This diff is at https://phab.parse.com/D11251, which will most likely be landed after I'm gone. These docs also refer to the CLP-ACL interaction docs that are missing from the new docs, so it would be good if those got moved over before this is merged.


hramos:
See #99

PR #90, opened 2015-06-10
Document supported API version

hallucinogen:
@grantland do we want to document this. This would be similar to https://parse.com/docs/ios/guide


grantland:
This is fine for now. We should probably put it in a more visible place like it's own paragraph in bold or in the sidebar. /cc @nlutsenko

nlutsenko:
This looks good, but I agree with @grantland that we might want to split this into a separate paragraph or highlight it somehow. I think the whole target experience is to make it easy to find, not something that people just stumble upon.

hallucinogen:
I'll do this for now. We'll change it for all SDK later if we have something in mind. At least now it's consistent across all of the SDKs

PR #91, opened 2015-06-10
Update iOS quickstart URL

hallucinogen:
Update the quickstart to be more precise.

@nlutsenko might want to take a look


nlutsenko:
LGTM

PR #94, opened 2015-06-11
Facebook users

bmbrina:
Minor details about using the Facebook SDK


nlutsenko:
Good catch! Will merge in after nit is fixed :D

PR #103, opened 2015-06-19
Document regular expressions

hramos:
Document startsWith/$regex, and include warning against using this over a large dataset. Regular expressions are discouraged, but are nonetheless part of the API.


gfosco:
Sorry, merged some other things and now this needs a rebase, otherwise it LGTM.

PR #108, opened 2015-06-20
(minor) Removes typo and Clarifies text

kayalew:
1. removes an unnecessary "of" (Typo) and clarifies language

2. removes a possible misconception that a queries match all items in a list


hramos:
Thanks for adding the clarification!

kayalew:
No probelm! Parse is awesome and this feels subjective so I wanted to state my reasons for changing text I felt was slightly awkward while reading the guide for the first time.

PR #113, opened 2015-06-25
Add a warning where override the constructor.

fcrosfly:
After I override the constructor and change the params. I get a strange error when I trying to save an object with a unsaved object as it's property

I found the following Code in ParseObject.php L493 which is used for create objects.

``` php
    public static function create($className, $objectId = null,
        $isPointer = false
    ) {
        if (isset(self::$registeredSubclasses[$className])) {
            return new self::$registeredSubclasses[$className](
                $className, $objectId, $isPointer
            );
        } else {
            return new ParseObject($className, $objectId, $isPointer);
        }
    }

```

Instead of suggest creating an static function create in the Subclass, I suggest adding a warning in the document.

Thank you.


gfosco:
Thanks. :+1:

PR #115, opened 2015-06-26
Update quick-reference.mdown

robertcastle:
Add example to Calling from Client Apps section.
Example shows how that the "X-Parse-REST-API-Key" needs to be changed as well as the value.


hramos:
I believe you can simply use the X-Parse-Client-Key header, even when calling from .NET.

robertcastle:
Yes you can. I was trying to make it clear that the X-Parse-REST-API-Key needs to be changed to the appropriate platform key or to X-Parse-Client-Key. Also to explicitly state what these keys are, as I could not find them mentioned in the docs.

hramos:
I opened #122 to address the key documentation issue. Can you update this PR to recommend using x-parse-client-key with the client key? No need to call out .NET specifically.

PR #120, opened 2015-07-07
Updated sample code

lewis-smith:
The sample code gives the following error:

Error finding related comments 102: pointer field post needs a pointer value

I guess this is something that has changed in Parse.


hramos:
Hmm, if "post" is a pointer, then the comparison should be against `request.object` instead of `request.object.id`. No need to create a new pointer to the Post as it's already present in `request.object`.

gfosco:
Committed a fix based on Héctors comment.

PR #125, opened 2015-07-19
Add JS code samples for Many-to-Many - Using Parse Relations

thewheat:


en/common/relations.mdown
```diff
@@ -394,8 +394,23 @@ book.saveInBackground()
 // Want to contribute to this doc? https://github.com/ParsePlatform/Docs
 ```
 ```common-js
-// no js example
-// Want to contribute to this doc? https://github.com/ParsePlatform/Docs
+// let’s say we have a few objects representing Author objects
+var authorOne =
```
hramos: Can you add ... to each of these to maintain consistency?

thewheat: Added accordingly

PR #128, opened 2015-07-21
Missing { in python example for modifying a schema

Egnaro:
The python example fails as it's missing the opening { in the json.


hramos:
It looks like this conflicts with one of your earlier PRs. Can you fix it?

gfosco:
Fixed.

PR #134, opened 2015-07-25
Replaces deprecated `refresh` method on PFObject for `fetch`

natanrolnik:


nlutsenko:
Nice catch. Thanks.

natanrolnik:
😃 👍

PR #145, opened 2015-07-31
Move and Add Sample Code for Join Table Section

thewheat:
One code sample was in the wrong location. Moved it accordingly and
added one missing sample code


hramos:
Looks like this may be conflicting with one of your other PRs I just merged. Can you fix the conflict?

thewheat:
My changes are no longer relevant as https://github.com/ParsePlatform/Docs/pull/144/ fixes the changes I was trying to make

PR #162, opened 2015-08-11
Fix app id tokens.

hramos:
Apps guide was missing the application id token. Updated results example to use a generic app as an example.


html5cat:
:+1: :shipit:

PR #163, opened 2015-08-11
Patch 1

pietrodegrazia:


html5cat:
@pietrodegrazia good catch! It's actually seconds, but the examples we provide are wrong. I'll update that.

PR #165, opened 2015-08-11
Organize error codes into sections

hramos:
Organize error codes into related groups, and add resolution steps for operational issues.

Add httpRequest connection timeout to Limits section.


html5cat:
:+1:

PR #168, opened 2015-08-12
Document concurrent_http_request_limit value

hramos:
Max number of concurrent outbound HTTP requests per V8 request is 8.


html5cat:
:+1: :bow:

PR #169, opened 2015-08-13
Role: Fix code 111 error

amowu:
If use `X-Parse-REST-API-Key` to create or update Role users relation, that can get `can't add a non-pointer to a relation` error.


html5cat:
@amowu that's not required for this operation. Let me know if you need help troubleshooting. Thanks!

PR #184, opened 2015-08-21
Add missing Swift placeholder section.

hramos:
Some code samples are missing a Swift equivalent. If there is no placeholder text, the entire code sample disappears when Swift is selected as the language.

This PR should take care of the disappearing code samples when switching to Swift. We still need actual code in these boxes.


hramos:
Skipping review due to the nature of this PR.

PR #185, opened 2015-08-21
Update saveInBackground methods for notifications

nickoneill:


nickoneill:
There are a ton of these old `saveInBackground` methods that need to be converted to `saveInBackgroudnWithBlock`. I'll PR them as I see them.

hramos:
As someone who always uses blocks, I haven't used the block-less inBackground methods recently. Are these no longer part of the SDK? IIRC, they should behave in a similar manner as the inBackgroundWithBlock method when a nil block is used.

nickoneill:
Nope, these no longer exist in the SDK so copy/paste from the docs gives build errors.

nlutsenko:
Ehm. These quite exist in the SDK and are available throughout.
If you are using Swift - please keep in mind that until 1.8.2 is out - you would need to add `import Bolts` to see them, since Swift preprocessor does some stupid things.

nlutsenko:
@nickoneill Example: https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/PFObject.h#L306

nickoneill:
Oh, sorry. It seems like it makes more sense to use the calls in the docs that work out of the box though, yeah?

nlutsenko:
True, I absolutely agree with you.
That's exactly why we found a way to fix it without requiring to import Bolts and those changes were already landed in master of the SDK and will be out shortly with v1.8.2

nickoneill:
Sounds good, I'll close this.

hramos:
Thanks for bringing this up anyway, @nickoneill. Having samples that work out of the box is something we're aiming for. As 1.8.2 should fix this shortly, this specific PR will not be necessary, but we would still appreciate PRs for any other issues you may run into while using the stock sample code.

PR #201, opened 2015-09-10
Update objects.mdown

javierdelacueva:
Update the docs to make clear that set(null) is not recommended instead use unset.


ericnakagawa:
Closing for PR with localized text.

PR #206, opened 2015-09-12
Update sample code in Relational Queries for Swift 2

cnmcgrath:
Fixed document to say PFObject instead of AnyObject


nlutsenko:
Awesome! Thanks!

PR #208, opened 2015-09-14
Special properties are not exposed to set and get methods

valgaze:
Added clarification that objectId, updatedAt, and createdAt cannot be modified with .set nor retrieved with .get. 

This has been a source of confusion for some: https://parse.com/questions/javascript-get-objectid


gfosco:
:+1:

PR #212, opened 2015-09-17
Count for queries should be Int32 in swift.

richardjrossiii:
It needs to be this way for legacy purposes mostly, but this would cause compile issues for people using our SDK in swift on 64-bit targets.


nlutsenko:
LGTM

PR #214, opened 2015-09-20
Synchronize accountkey command example

adamgross42:
The documentation states `parse configure accountkey` but the example uses `parse configure key` - this changes the latter to the former, because that is also what's used in the Parse CLI messages.


nlutsenko:
cc @pavanka 
LGTM though, merging now.

PR #215, opened 2015-09-22
Improving @javierdelacueva's fix and localizing.

ericnakagawa:
This informs developers not to use object.set(null) for removing fields from objects. (t8321814)


html5cat:
@ericnakagawa looks good pending https://github.com/ericnakagawa/Docs/pull/1

html5cat:
:shipit:

PR #223, opened 2015-10-05
Swift 2 println changes

Bandd-k:
I've changed every "println" to "print".


nlutsenko:
Perfect! Thanks for fixing these!
Can you squash all these commits into a single one? Will merge straight after.

nlutsenko:
Superseeded by #224

PR #234, opened 2015-10-14
Updating to use new version of swift syntax

ltcompounder:
Saw this error while trying to implement parse onto my xcode file. When creating new user.

### DIFF
--- a/en/ios/users.mdown
+++ b/en/ios/users.mdown
@@ -48,7 +48,7 @@ func myMethod() {
   user["phone"] = "415-392-0202"
 
   user.signUpInBackgroundWithBlock {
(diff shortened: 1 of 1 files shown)


nlutsenko:
Your change is absolutely valid Swift syntax, but turns out the old one works as well and gives absolutely 0 errors with the latest SDK.

Used this snippet to try to reproduce an error:
```swift
PFUser().signUpInBackgroundWithBlock { (succeeded: Bool, error: NSError?) -> Void in
  print("success: \(succeeded), error: \(error)")
}
```

nlutsenko:
The change purely makes the type annotations be implicit, instead of being explicit.
Where the old one is probably more helpful in docs, since it's explicit and directly specifies the type.

PR #236, opened 2015-10-22
Update queries.down - specifically, Querying Subclasses

dudgeon:
I feel like the whole point of subclassing is to treat our PFClasses like first class types in Swift; therefore, why not cast the returned results as the specific PFObject type they are? In this case 'Armor' class.


nlutsenko:
Nice catch! Thanks!

PR #242, opened 2015-10-29
Added documentation queries on arrays

rubencodes:
Specifically, for finding an object where a key's array-type value contains one or more elements of an array.


hramos:
It looks like containedIn is already documented earlier in the file, although I do see the value of surfacing it again under the array section.

Could you perhaps update the copy to be closer to what was used earlier in the document (and I quote):

    If you want to retrieve objects matching any of the values in a list of values, you can use containedIn, providing an array of acceptable values.

Finally, since this section is repeated in each of the platform guides, it would be great if this pull request made the exact same change to each of the platforms for consistency.

PR #269, opened 2015-11-22
Codefix - ReferenceError: SendGrid is not defined

sagans:
In your Parse documentation.
https://parse.com/docs/cloudcode/guide#cloud-code-modules-sendgrid

Search for "SendGrid.sendEmail("

(I think) that should be a lowercase "S", rewritten as "sendGrid.sendEmail("

Thats it...
Sagan


hramos:
Good find! Looks like the proposed fix does not match the original sendgrid variable name, however.

sagans:
That was a sad (embarrassing) case of a 5 and 9 year old jumping on my back while making a code fix!

hramos:
LOL

PR #273, opened 2015-11-24
Add notes regarding built-in Cloud Module versions.

hramos:
Parse's built-in Cloud Modules are not versioned, and therefore have not been updated since their introduction. This means that customers interested in taking advantage of the APIs in Moment.js v2 would need to import the newer library into their Cloud Code instead of using the built-in Moment module. The same goes for Stripe, Twilio, and other libraries.

This PR points users to the newer libraries.


ericnakagawa:
LGTM

hramos:
TY

PR #290, opened 2015-12-27
beforeSave does not update modified object

sebastianmach:
When modifying an object in the beforeSave trigger, it needs to be specified when calling response.success(). Otherwise parse does not notice that the object has been modefied.
This error occoured when running the cloud code local or on heroku (maybe it does not happen when running the code on the parse cloud code platform).

When calling `response.success();`, the HTTP response looks like this (even if the object has been modified):

```
{
    "success": true
}
```

Whenn calling `response.success(request.object);`, the response looks something like this:

```
{
    "success": {
        ** the modified object here **
    }
}
```

When returning the modified object, parse will save the modified object.

### DIFF
--- a/en/cloudcode/cloud-code.mdown
+++ b/en/cloudcode/cloud-code.mdown
@@ -193,7 +193,7 @@ Parse.Cloud.beforeSave("Review", function(request, response) {
     // Truncate and add a ...
     request.object.set("comment", comment.substring(0, 137) + "...");
   }
-  response.success();
+  response.success(request.object);
 });
 ```
 


hramos:
This is only necessary when using Webhooks AFAICT. This is documented as such under the [beforeSave webhooks](https://parse.com/docs/cloudcode/guide#cloud-code-advanced-beforesave-webhooks) documentation. I'm inclined towards using the minimum sample code that works in Cloud Code, but I would consider merging this in if this turns out to cause sufficient confusion as more developers adopt webhooks.

electronicMe:
Thank you for your fast reply!

To tell you more about my test scenario, I used the the parse CLI to create a Parse App Backend and deploy it on heroku. I'm using the standard node.js APIs from parse.

When using Webhooks, it is definitely required to specify the modified object in the `response.success()` call because parse obviously can't detect the changes of the object (as mentioned in my original comment).
When I first tried to modify an object, I didn't know that the beforeSave method call actually comes from the Express.js framework. I thought returning the modified object is automatically done by the parse API.
However, when running the cloud code on parse servers it might not be necessity to specify the modified object in the `response.success()` call. (I didn't test that and I don't know how the parse servers react in this case).

But when reading the "MODIFYING OBJECTS ON SAVE" article [here](https://parse.com/docs/cloudcode/guide#cloud-code-modifying-objects-on-save) the text says:

> You just call `response.success` on the altered object.

what is not really accurate because you neither call a `response.success` method on the altered object (which is in this case `request.object`) nor you specify the altered object when calling the `response.success` method.

After solving my problem, I thought the documentation text meant you should specify the altered object `response.success` method (which is not documented in the example).

PR #291, opened 2015-12-28
fix en/ios/objects.mdown subtitle

limaofuyuanzhang:
The last three parts are part of subclass, use title of three level is
more appropriate.


hramos:
Thanks! Good catch.

Issue #136, opened 2015-07-27
Update examples to use Swift 2.0


hramos:
Notes:

- println is now print

PR #292, opened 2015-12-29
Updated UIUserNotificationType and println to swift 2 syntax

lukeharback:
I have changed some syntax and wanted to explain my decision for line 26 in the file en/ios/push-notifications.mdown

For it to build without errors in swift 2, the '|' operator is not used. Instead it is an array of UIUserNotificationType's that gets set into one UIUserNotificationType. This can not be inferred, so it is important to define the type of the constant. Now defined, I thought we could drop the 'UIUserNotificationType' before every value of the enum we specify. What are the thoughts on this? @hramos does this help with #136

### DIFF
--- a/en/ios/push-notifications.mdown
+++ b/en/ios/push-notifications.mdown
@@ -23,7 +23,7 @@ UIUserNotificationSettings *settings = [UIUserNotificationSettings settingsForTy
 [application registerForRemoteNotifications];
 ```
 ```swift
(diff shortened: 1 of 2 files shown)


nlutsenko:
Hey @lukeharback, thanks for making this happen.
Dropping `UIUserNotificationType` before enum case in favor of explicitly typing the variable - absolutely!
And it does indeed help with #136.

lukeharback:
@nlutsenko Awesome, glad I could help!

PR #294, opened 2016-01-03
Update queries.mdown [Correct way to query now]

gustavoloureirodosreis:
None of these queries work anymore. I've changed the first one to show how it works now. Someone, please, change the rest.


hramos:
Hey @Gustavo23 thanks for the initiative. May I suggest opening an Issue with this example instead? As it stands, this PR cannot be merged until all the query examples have been updated to maintain consistency.

PR #300, opened 2016-01-06
UIBackgroundModeNoData not a valid type.

stephen-francis:
UIBackgroundModeNoData is not a type of UIBackgroundFetchResult, this should be UIBackgroundFetchResultNoData.


nlutsenko:
Nice catch! Thanks!

PR #304, opened 2016-01-08
Push notifications docs have misleading parameter name

sceeter89:
Actually one should set `push_time` parameter not `timeZone`.


hramos:
Good catch!

PR #307, opened 2016-01-15
Update todo_view to use .png instead of .gif

hallucinogen:
The image is not animated anyway.

@nlutsenko @hramos


hramos:
Huh, this used to be an animated GIF that showed scrolling.

LGTM

PR #310, opened 2016-01-18
Replace chained bitwise operators

huntermonk:
Swift is no longer supporting chaining bitwise operators. Instead, you need to pass an array of PFLogInFields.


nlutsenko:
Nice catch, thanks!

huntermonk:
No problem. Thanks for being responsive. That was my first public pull request 🙌🏼

PR #317, opened 2016-01-24
Fix ios/push-notifications.mdown

knchst:
"@" was forgotton.

modified:   cn/ios/push-notifications.mdown
modified:   en/ios/push-notifications.mdown
modified:   jp/ios/push-notifications.mdown
modified:   kr/ios/push_notifications.mdown
modified:   ru/ios/push-notifications.mdown
modified:   tw/ios/push-notifications.mdown


nlutsenko:
Nice catch! Thanks!

PR #318, opened 2016-01-25
Remove references to WatchKit from Extensions guide.

nlutsenko:
This doesn't work on watchOS 2 and might be confusing.


hramos:
LGTM

PR #325, opened 2016-01-29
Updating deployment.md

pivotal-mikegresham:
Adding instructions for deploying the server to Pivotal Web Services (http://run.pivotal.io).


hramos:
Hi,

We're not currently adding any more providers to this list, but rather we're keeping a list at https://github.com/ParsePlatform/parse-server/wiki. Can you write this up and host it on a blog off-site so we can link to it?

PR #327, opened 2016-01-29
Add instructions to deploy on Clever Cloud

BlackYoup:
Hi guys, this adds an how-to to deploy on the [Clever Cloud platform](https://www.clever-cloud.com/en/)


hramos:
Thanks for writing this up. We're currently not hosting vendor-specific guides in the Server Guide. If you can write this up as a blog post off site, it can be added to the list at https://github.com/ParsePlatform/parse-server/wiki

PR #342, opened 2016-02-16
add mongorocks quickstart guide

tredman:


drew-gross:
Nice!

hramos:
@tredman This content looks great. Can you add it to the Parse-Server wiki instead?

We are now redirecting parse.com/docs/server/guide to https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide. We should actually remove en/server/**.* from the repo.

drew-gross:
@hramos parse.com/docs has much better navigation and stuff. Is it possible to get the nice navigation while still having the docs in a wiki?

hramos:
@drew-gross I also like the navigation and design that parse.com/docs provides, but right now the wiki in the GitHub repo for Parse Server is the source of truth. We should focus on getting the README and Wiki over there polished now. We can revisit hosting these docs again on parse.com at a later date.

tredman:
Done.

PR #344, opened 2016-03-02
Modifies a typo in the afterDelete trigger example

DohanKim:
'.id' was missing.

### DIFF
--- a/en/cloudcode/cloud-code.mdown
+++ b/en/cloudcode/cloud-code.mdown
@@ -259,7 +259,7 @@ In some cases, you may want to perform some action, such as a push, after an obj
 ```js
 Parse.Cloud.afterDelete("Post", function(request) {
   query = new Parse.Query("Comment");
-  query.equalTo("post", request.object);
+  query.equalTo("post", request.object.id);
   query.find({
     success: function(comments) {
       Parse.Object.destroyAll(comments, {


nlutsenko:
It actually will work without an `id`, and will do simple pointer matching. 
I might be missing something but the model here suggests that if you have a `Post` object and a `Comment` object and every `Comment` has a pointer to `Post` - after deletion of `Post` you delete all of it's comments. Not sure if this change is necessary if that's the case.

@DohanKim, what are your thoughts on this?

DohanKim:
@nlutsenko 

according to explanation of 'equalTo' function in the API Reference, 
( https://www.parse.com/docs/js/api/classes/Parse.Query.html#methods_equalTo )
the second argument have to be a Post object's ID (request.object.id) rather than a javascript pointer (request.object), as it is the value of Comment objects' foreign key indicating specific post.


------------------------------------------------------
Parse.Query equalTo( key, value ) 
Adds a constraint to the query that requires a particular key's value to be equal to the provided value.

Parameters:
key <String>
The key to check.

value <Object>
The value that the Parse.Object must contain.

hramos:
Agree with @nlutsenko. This example assumes that "post" is a Pointer. A comparison by object id (a String) would fail here.

DohanKim:
beforeDelete Trigger's example (https://www.parse.com/docs/cloudcode/guide#cloud-code-beforedelete- triggers) is showing exactly opposite. @hramos, then either one of example code of afterDelete and beforeDelete should be wrong.



```js
query = new Parse.Query("Photo");
query.equalTo("album", request.object.id);
```

hramos:
@DohanKim not necessarily. They're two different classes and different columns. One could have a Pointer, the other a String.

PR #345, opened 2016-03-15
Update docs to reflect the preferred async nature of logOut

andrewimm:
`logOut` should be treated as async, since we introduced Sessions in April 2014.


nlutsenko:
Please merge!

PR #347, opened 2016-03-18
adding live query docs

peterdotjs:


en/js/live-queries.mdown
```diff
@@ -0,0 +1,236 @@
...
+Parse.LiveQuery.close();
+```
+When you are done using LiveQuery, you can call `Parse.LiveQuery.close()`. This function will close the WebSocket connection to the LiveQuery server, cancel the auto reconnect, and unsubscribe all subscriptions based on it. If you call `query.subscribe()` after this, we will create a new WebSocket connection to the LiveQuery server.
+
+
+## Setup server URL
+```javascript
+Parse.liveQueryServerURL = 'ws://XXXX'
+```
+Most of the time you do not need to manually set this. If you have setup your `Parse.serverURL`, we will try to extract the hostname and use `ws://hostname` as the default `liveQueryServerURL`. However, if you want to define your own `liveQueryServerURL` or use a different protocol such as `wss`, you should set it by yourselves.
```
drew-gross: s/yourselves/yourself

en/js/live-queries.mdown
```diff
@@ -0,0 +1,236 @@
...
+When some network error or LiveQuery server error happens, you'll get this event.
+
+<br/>
+
+***
+## Advanced API
+In our standard API, we manage a global WebSocket connection for you, which is suitable for most cases. However, for some cases, like when you have multiple LiveQuery servers and want to connect to all of them, a single WebSocket connection isn't enough. We've exposed the `LiveQueryClient` for these scenarios.
+
+## LiveQueryClient
+A `LiveQueryClient` is a wrapper of a standard WebSocket client. We added several useful methods to help you connect/disconnect to LiveQueryServer and subscribe/unsubscribe a `ParseQuery` easily.
```
wangmengyan95: s/added/add

peterdotjs:
Fixed per comments.

PR #356, opened 2016-04-26
Translation from English to Brazilian Portuguese of the Parse Server, Common Errors and Common data

felipemobile:
This is the first step to full translate Parse docs to Brazilian Portuguese. As I’m not a professional translator there are some mistakes to be fixed yet.

### DIFF
--- a/pt/android/config.mdown
+++ b/pt/android/config.mdown
@@ -1,0 +1,80 @@
+#Configuração 
+
+## Configuração do Parse
+
+`ParseConfig` é a maneira de configurar sua aplicação remotamente, armazenando um único objeto de configuração do Parse. Ele permite que você adicione
+em sua aplicação algo como uma simples 'Mensagem do Dia'. Para começar usando o `ParseConfig`, você precisa
+adicionar alguns parâmetros (chave/valor) em sua aplicação através da Dashboard.
+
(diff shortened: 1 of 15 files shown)


hramos:
Thanks for working on this! We're actually going to take another pass at our docs soon to convert them to Parse Server specific docs, which unfortunately may mean that some of these translations will soon become stale.

My portuguese is very basic but I should be able to take a stab at reviewing this diff.

felipemobile:
@hramos I realized I've started translating on the wrong path :-)

No problem if you want to discard this diff. My interest on translations to Brazilian Portuguese is because I've been working on a Parse replica for our Brazilian customers [colubris.com.br](https://colubris.com.br).

Fortunately many of our apps rely on Parse's infrastructure, and now I've the opportunity to improve and add an extra layer on top of Parse's services.

felipemobile:
What do you think about iOS, Android, JS and REST translations?

hramos:
Yeah, the canonical server docs are actually in the @ParsePlatform/parse-server wiki now, and we don't support internationalization there. I'm planning on moving these back to this repo soon where we can again take advantage of this.

All of the client SDK guides will remain in this repo and will continue to power the docs at parse.com/docs.

PR #371, opened 2016-08-25
Update getting-started.md

kwstasna:
Mentioned how to initialize parse-server with Javascript.

### DIFF
--- a/_includes/js/getting-started.md
+++ b/_includes/js/getting-started.md
@@ -12,7 +12,7 @@ Our JavaScript SDK is originally based on the popular [Backbone.js](http://backb
 
 Our SDK supports Firefox 23+, Chrome 17+, Safari 5+, and IE 10. IE 9 is supported only for apps that are hosted with HTTPS.
 
-To initialize your own Parse-Server with Javascript, you should replace your curently initialize code with this
+To initialize your own Parse-Server with Javascript, you should replace your current initialization code with this
 <pre><code class="javascript">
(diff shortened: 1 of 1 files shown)


_includes/js/getting-started.md
```diff
@@ -11,3 +11,9 @@ The Parse platform provides a complete backend solution for your mobile applicat
 Our JavaScript SDK is originally based on the popular [Backbone.js](http://backbonejs.org/) framework, but it provides flexible APIs that allow it to be paired with your favorite JS libraries. Our goal is to minimize configuration and let you quickly start building your JavaScript and HTML5 app on Parse.
 
 Our SDK supports Firefox 23+, Chrome 17+, Safari 5+, and IE 10. IE 9 is supported only for apps that are hosted with HTTPS.
+
+To initialize your own Parse-Server with Javascript, you should replace your curently initialize code with this
```
hramos: Typo in 'curently'. Maybe use 'current initialization'.

_includes/js/getting-started.md
```diff
@@ -11,3 +11,9 @@ The Parse platform provides a complete backend solution for your mobile applicat
...
+<pre><code class="javascript">
```
hramos: I believe you can use triple backticks here. The format you're using is only necessary in cases where the included code has &lt; or &gt; which sometimes messes up the renderer.

kwstasna:
I believe it's ok now.
Actually it should be updated also for all the other SDK's too.

PR #372, opened 2016-08-27
Adds documentation for Cloud Jobs

flovilmart:


flovilmart:
@drew-gross: regarding https://github.com/ParsePlatform/parse-server/pull/2560

hramos:
Content LGTM. I think we did use to have a cloud jobs section that I deleted as the documentation was specific to the hosted Parse.com version, perhaps it's worth it taking a look through the history to see if there's any sample content worth resurfacing here.

 Was the bundle.js change intentional? Looks like mostly whitespace/formatting changes.

flovilmart:
@hramos yep, we should take a look, also the Dashboard will have Jobs feature compatible with parse-server (view and run manually)

As for the bundle, those changes are not intentional at all, I thought that was generated code :)

PR #377, opened 2016-09-18
[WIP] Parse server wiki

flovilmart:
- Merges the parse-server wiki history into that repository to preserve files commiters
- Adds the most important parts of the ParseServer guide wiki here.


hramos:
This is an awesome PR.

I think there's two separate guides:
- The Parse Server guide - how to configure and deploy your own Parse Server
- The Parse Server Developer guide - how to make code level changes and expand Parse Server as a contributor

I feel like the latter would fit better in the parse-server GitHub repo. If you think that splitting the docs across two locations like this would be confusing, maybe we can split the docs into two guides in this repo?

flovilmart:
Yep, the later definitely goes into the wiki/parse-server repo, that's the goal, there are still a lot of things to adjust before it's mergeable.

I don't believe it would be confusing, a link on the wiki of parse-server + readme would be nice. We could keep the parse-server README tight and easy.

flovilmart:
Hot!

PR #387, opened 2016-11-21
Update syntax for loginInBackground example

jaimeagudo:


flovilmart:
The existing code should correctly work in swift. Why that change?

nitrag:
@jaimeagudo Is correct. I would go as far as to include the error parsing, which was quite difficult to figure out on my own.

```
PFFacebookUtils.logInInBackground(withReadPermissions: ["public_profile","email"]) {
            (user: PFUser?, error: Error?) -> Void in
            if let error = error {
                if let errorString = (error as NSError).userInfo["error"] as? String {
                    NSLog(errorString);
                }
            } else {
                if let user = user {
                    if user.isNew {
                        print("User signed up and logged in through Facebook!")
                    } else {
                        print("User logged in through Facebook!")
                    }
                    self.currentUser = PFUser.current();
                    self.loggedIn()
                }
            }
        }
```

_includes/ios/users.md
```diff
@@ -545,10 +545,10 @@ There are two main ways to use Facebook with your Parse users: (1) to log in (or
 }];
 </code></pre>
 <pre><code class="swift">
-PFFacebookUtils.logInInBackgroundWithReadPermissions(permissions) {
-  (user: PFUser?, error: NSError?) -> Void in
-  if let user = user {
-    if user.isNew {
+PFFacebookUtils.logInInBackground(withReadPermissions: permissions) {
+  (user: PFUser?, error: Error?) in
+  if let u = user {
```
flovilmart, line 550: revert that change please

_includes/ios/users.md
```diff
@@ -545,10 +545,10 @@ There are two main ways to use Facebook with your Parse users: (1) to log in (or
...
+    if u.isNew {
```
flovilmart, line 551: that one too.

flovilmart (changes requested):

_includes/ios/users.md
jaimeagudo: That's the current syntax and _Intellisense_ in Xcode is *&^#, why do you want to revert it?

flovilmart: we prefer `let user = user` usually

addisonElliott:
Was looking through the code, and I fail to see where logInInBackground is defined for PFFacebookUtils? Where is the code at?

flovilmart:
@addisonElliott https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/ParseFacebookUtils/ParseFacebookUtils/PFFacebookUtils.h#L74

TomWFox:
As this is dormant I have merged #589 which fixes the syntax as suggested here with respect to the suggested changes from Flovilmart.

PR #389, opened 2016-11-25
Update Init for Facebook

nitrag:
Updating for swift 3

### DIFF
--- a/en/ios/users.mdown
+++ b/en/ios/users.mdown
@@ -486,7 +486,7 @@ import FBSDKCoreKit
 import ParseFacebookUtilsV4
 
 // AppDelegate.swift (Swift 3)
-func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -&gt; Bool {
+func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
   
   //your other app delegate stuff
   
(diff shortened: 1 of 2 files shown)


nitrag:
Signed CLA.

en/ios/users.mdown
```diff
@@ -485,10 +485,19 @@ There's also two code changes you'll need to make. First, add the following to y
 import FBSDKCoreKit
 import ParseFacebookUtilsV4
 
-// AppDelegate.swift
-func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
-  Parse.setApplicationId("parseAppId", clientKey:"parseClientKey")
-  PFFacebookUtils.initializeFacebookWithApplicationLaunchOptions(launchOptions)
+// AppDelegate.swift (Swift 3)
+func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -&gt; Bool {
```
hramos: Is \&gt\; necessary here? It should render fine within the code block, and it looks like you didn't need to use this encoding later in the same file.

hramos (changes requested):

TomWFox:
I'm sorry that this has been over looked. Would you be able to update this code for swift 4.2 (+ the latest version of the Parse iOS SDK) and then I will take another look?

Also would you be able to remove the vertical parameter alignment as it doesn't appear well on the docs:
screenshot 2019-03-08 at 11 58 12

TomWFox:
I'm closing this now as I have started a new pull request (#594) that brings in these changes (those of which are still needed).

PR #405, opened 2017-02-03
inform of deprecation on parse-server

jaimeagudo:
Avoid newcomers wasting time implementing something useless
See https://github.com/ParsePlatform/parse-server/issues/3455

### DIFF
--- a/_includes/ios/push-notifications.md
+++ b/_includes/ios/push-notifications.md
@@ -1,5 +1,16 @@
 # Push Notifications
 
+
+
+
+**DEPRECATED FOR PRODUCTION ON PARSE SERVER ** See http://blog.parse.com/learn/engineering/the-dangerous-world-of-client-push/ and perhaps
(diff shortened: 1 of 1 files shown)


hramos:
Instead of adding a warning, let's remove the content that is no longer relevant on Parse Server.

The content on setting up push and using installations is still relevant regardless of whether clients can send pushes or not, so let's leave that in.

Can you update your PR to ensure only the outdated sections are removed? Ideally this PR would touch all other platforms as well as this is not iOS-specific.

jaimeagudo:
You are right but at the moment I don't have the time to go through this, but, IMHO is better to merge this and add a TODO than leaving it as it's

hramos:
What I am advocating for is removing the stale content. I'd rather prune the no longer relevant sections and not add a warning. Seems like an equal amount of work for doing the right thing.

TomWFox:
I'm closing this now as I have opened a PR which overhauls the push notifications section - #604.

PR #406, opened 2017-02-10
Switching from oauth to auth to line up with https://github.com/ParsePlatform/parse-server/issues/3175

AndrewLane:
Switching from oauth to auth to line up with https://github.com/ParsePlatform/parse-server/issues/3175


flovilmart:
Thanks!

PR #407, opened 2017-02-12
Updates PHP Refs

montymxb:
- Updates the ```ParseClient::setServerURL``` reference, which now takes the server url and mount path separately. 
- Updates all ```ParsePush::send``` calls to use the master key, which is required when working with parse server.  
- Also updates info regarding getting the status of recent pushes and adds use cases for reading PushStatus from recent push requests via ```ParsePush::hasStatus``` and ```ParsePush::getStatus``` (which return an instance of the new ```ParsePushStatus```). This was recently added to the PHP sdk for reading the volatile class ```_PushStatus```.


montymxb:
@hramos anything else that should be done for this to be given a look over? I haven't bothered but let me know if you need these squashed.

hramos:
Looks good at a high level but I'd defer to someone more knowledgeable in the PHP SDK. Who added the Push Status changes? Can they review this?

montymxb:
@hramos I did actually, the original PR is [here](https://github.com/ParsePlatform/parse-php-sdk/pull/283) for reference. 

@gfosco did briefly provide some input to the process. Perhaps he (if he has the time) or another parse member familiar with the php sdk would be willing to check this and give their opinion?

hramos:
Works for me.

montymxb:
👍 thanks!

PR #409, opened 2017-02-14
Update algolia apiKey

QuentinDela:


flovilmart (approved):

hramos:
How was this key generated, and is it tied to a paid plan?

QuentinDela:
I asked for one on support@algolia.com

hramos:
Thanks Quentin, I chatted with them and they let me know. They reactivated the original key so we're good now.

flovilmart:
So? Revert?

hramos:
@flovilmart already happened :)

PR #421, opened 2017-03-28
Add required rest-api-key on curl sample

jaimeagudo:


flovilmart:
The REST key should not be set here as the master key is passed. Can you explain?

flovilmart:
@jaimeagudo ?

jaimeagudo:
Hi @flovilmart, sorry for the late reply. I guess you are right as per the current version, I recall it was required (perhaps due some local tweaks, not sure) when I issued the PR. Since then I upgraded the server and you are right cause now my test doesn't require the REST key

Sorry for distracting you

flovilmart:
Uhm, that's good to know, we perhaps had a bug, in any case, thanks for contributing back :)

PR #429, opened 2017-04-21
Updated MongoDB Prerequisites

a-harrison:
Added MongoDB 3.2 as an acceptable version prerequisite. This is an acceptable version per http://docs.parseplatform.org/parse-server/guide/#database; version supported verified in https://github.com/parse-community/parse-server/issues/1867.


flovilmart:
Thanks for pointing that out!

PR #431, opened 2017-05-02
Add information about different redis databases

stefanraffeiner:
Both, LiveQuery and RedisCacheAdapter use a redisURL. As the CacheAdapter might flush the database, it's recommended to use two different databases. Added this information to the docs.
This is also important if you run additional services like kue on redis.

Thanks for the help @JeremyPlease


flovilmart (approved):

flovilmart:
Thanks!

PR #435, opened 2017-05-12
Update push notifications docs

funkenstrahlen:
The parse push adapter now supports sending push notifications to macOS and tvOS. I updated the docs to explain how the configuration works.


flovilmart:
Thanks!

PR #436, opened 2017-05-20
SERVER_URL to show glitch.me

ghcodebase:
HyperDev is now using the branding and domain Glitch.  I spent about five hours today working out that the server was in fact:
SERVER_URL=https://noun-noun.glitch.me/parse
not
SERVER_URL=https://noun-noun.hyperdev.space/parse
This might save others the time I lost.


flovilmart (approved):

flovilmart:
Thanks for opening that up! It's hard to keep up with everything :)

PR #439, opened 2017-06-01
Update security.md

pablolic:
Parse.Cloud.useMasterKey has been deprecated and converted to a no-op in Parse Server version 2.3.0,  I'm correcting this in this document as it is confusing users.


flovilmart (approved):
Thanks for the improvement!

PR #441, opened 2017-06-01
Update getting-started.md

itzharDev:
configure manifest application name  with application class

### DIFF
--- a/_includes/android/getting-started.md
+++ b/_includes/android/getting-started.md
@@ -64,3 +64,11 @@ public class App extends Application {
   }
 }
 ```
+
+- **Option 3:** Setup `Application` as your application name in `AndroidManifest.xml`
+```xml
+<application ...
+  android:name=".App">
+  
+</application>
+```


_includes/android/getting-started.md
```diff
@@ -64,3 +64,11 @@ public class App extends Application {
   }
 }
 ```
+
+- **Option 3:** Setup `Application` as your application name in `AndroidManifest.xml`
+```xml
+<application ...
+  android:name=".App">
```
hermanliang, line 71: I think this change can be integrated with `Option 1` on line 23, right?

hermanliang (changes requested):

itzharDev:
I think It's better to put it in the end because u need the class name...

hermanliang:
@itzharDev I know, but it is not an option to set parse. It more likely to be a next step (`Step 3`) of installation.

addisonElliott:
This can be closed since the similar code was merged in #468

_includes/android/getting-started.md
Unic-1: That would be fine

addisonElliott:
Bump. Can someone close this?

Issue #442, opened 2017-06-09
Relations Sections Defaulting to JS


montymxb:
A user brought up an [issue](https://github.com/parse-community/parse-php-sdk/issues/321) with the docs regarding the php sdk. After taking a look I noted that it appears _all_ the sdks exhibit the same issue. In such sections as 'Relations' the code samples appear to be defaulting to JS where there are multiple samples to choose from.

An example of this can be seen in the [php guide](http://docs.parseplatform.org/php/guide/#using-pointers), the [ios guide](http://docs.parseplatform.org/ios/guide/#using-pointers) and the [android guide](http://docs.parseplatform.org/android/guide/#using-pointers). It looks like the issue stems from shared files in the `_includes/common`, where multiple languages may be conditionally displayed with the same documentation, using expressions such as `{: .common-lang-block .php }`.

I'm not certain but taking a quick look this appears to be stemming within the javascript which dynamically displays the appropriate code sections. I'm not too familiar with this project but if no one else is sure I can take a look at rectifying this issue myself.

flovilmart:
It seems to be all good! Great job!

montymxb:
Excellent! Glad to hear it

PR #443, opened 2017-06-09
Fix for invalid 'platform'

montymxb:
This is a patch for the observed issue in #442, where common sections with multiple code samples would default to always showing JS no matter the current sdk.

Originally, before the docs were at docs.parseplatform.org they used to be at parseplatform.org/docs (if I believe correctly). When determining which set of code snippets to display the JS pulled the platform 'name' from the url as follows
```js
var platform = window.location.pathname.split('/')[2];
```
Now that docs are served using the **docs** subdomain instead of the docs directory this retrieves the directory 1 _past_ the intended platform name. The new result was just returning `guide` for every page. Since this is not a valid platform the 'common' language block to show would fall back to the default case, which also happens to be the JS case.
```js
case 'js':
default:
    $('.common-lang-block.js').show();
```
The fix is just a one liner accounting for the now absent `docs` directory, simply changing the index of the split pathname to 1 instead of 2.

### DIFF
--- a/assets/js/bundle.js
+++ b/assets/js/bundle.js
@@ -801,7 +801,7 @@
 			hljs.highlightBlock(block);
 	});
 
-	var platform = window.location.pathname.split('/')[2];
+	var platform = window.location.pathname.split('/')[1];
 	if (platform) {
 			new App.Views.Docs.Main({
 					language: 'en',


flovilmart (approved):

flovilmart:
Let's see if it's all 😊

PR #445, opened 2017-06-18
Add docs about Parse.Query.and()

gofabian:
Update according to https://github.com/parse-community/Parse-SDK-JS/pull/367

### DIFF
--- a/_includes/js/queries.md
+++ b/_includes/js/queries.md
@@ -316,9 +316,17 @@ query.count({
 });
 </code></pre>
 
+
 ## Compound Queries
 
- If you want to find objects that match one of several queries, you can use `Parse.Query.or` method to construct a query that is an OR of the queries passed in.  For instance if you want to find players who either have a lot of wins or a few wins, you can do:
+For more complex queries you might need compound queries. A compound query is a logical combination (e. g. "and" or "or") of sub queries.
+
+Note that we do not support GeoPoint or non-filtering constraints (e.g. `near`, `withinGeoBox`, `limit`, `skip`, `ascending`/`descending`, `include`) in the subqueries of the compound query.
+
+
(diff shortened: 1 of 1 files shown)


natanrolnik (approved):

_includes/js/queries.md
```diff
@@ -328,16 +336,58 @@ var fewWins = new Parse.Query("Player");
...
+query.find()
+  .then(function(results) {
+    // results contains a list of users both older than 18 and having friends.
+  })
+  .catch(function(error) {
+    // There was an error.
+  });
+</code></pre>
+
+Sometimes world is complexer than this simple example and you may need an compound query of sub queries. You can use `Parse.Query.and` method to construct a query that is an AND of the queries passed in. For instance if you want to find users in the age of 16 or 18 who have either no friends or at least 2 friends, you can do:
```
addisonElliott, line 366: I would switch complexer to more complex. I'm not entirely sure complexer is a word, but I could be wrong.

Additionally, I think the first and second sentence can be morphed together to explain that the simple query above may not be sufficient in all cases and Parse.Query.and solves the problem.

Here is an example of what this could look like:

For more complex queries, you can use the `Parse.Query.and` method to return a query that is an AND of the input queries. For instance if you want to find users at the age of 16 or 18 who have either no friends or at least 2 friends, you can do:

addisonElliott (changes requested):
Some of the wording needs to be altered a bit. Altogether, I like the commit because I wasn't aware of the Parse.Query.and function myself.

addisonElliott:
What's the status on this PR?

natanrolnik:
I don't have write access here. @flovilmart ?

flovilmart:
Should be good now!

PR #446, opened 2017-06-24
Option 2 Missing Step

eduardo22i:
Register the App Class in the Android Manifest.

### DIFF
--- a/_includes/android/getting-started.md
+++ b/_includes/android/getting-started.md
@@ -64,3 +64,12 @@ public class App extends Application {
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


addisonElliott:
I like this commit but my suggestion is to change the language to indicate to the user that they need to add this piece of code regardless of if they choose option 1 or 2.

Something like this:
For either option, the custom `Application` class must be registered in `AndroidManifest.xml`:

addisonElliott:
Same thing solved in #468.

This can be closed now

montymxb:
Closing out, thanks again @addisonElliott !

PR #448, opened 2017-06-27
update cloud code doc to use promise instead of handlebar pattern.

acinader:
i happened to touch the file, so I updated to get away from the handlebar pattern and also introduce the .catch() backstop for promises that I wish I knew when I started....


flovilmart (approved):
Always nice to see some modernity!

PR #449, opened 2017-06-27
First pass at read preference

acinader:
1. just getting started, will need to be adjusted when we make client accessible
2. are there more examples I should add??

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -342,6 +342,14 @@ Parse.Cloud.beforeFind('MyObject', (req) =>  {
   // rejecting promise
   return Promise.reject('error');
 });
+
+// Setting the read preference for a query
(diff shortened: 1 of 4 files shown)


acinader:
@davimacedo please take a look.  anything we can add here??

davimacedo:
I think it is only missing the includeReadPreference option

assets/js/bundle.js
```diff
@@ -46,7 +46,7 @@
 /************************************************************************/
 /******/ ([
 /* 0 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
```
flovilmart: not sure this should be committed.

flovilmart (changes requested):

acinader:
@flovilmart

flovilmart (approved):

PR #454, opened 2017-07-15
Include reminder for a common mistake

billylo1:
CocoaPods new users often forget to use the newly generated xcworkspace, leading to a strange linker error that is hard to debug.


flovilmart (approved):

flovilmart:
Thank! Now it’s in!

PR #458, opened 2017-07-28
Changes references to api.parse.com and /1/

montymxb:
The current [REST Guide](http://docs.parseplatform.org/rest/guide/#quick-reference) and parts of the other guides contain numerous references to the now inactive `api.parse.com`, and the old `/1/` version 1 mount path. As it stands this is misdirecting some individuals, as seen in [#694](https://github.com/parse-community/Parse-SDK-Android/issues/694) over in the Android sdk. 

This proposes to change those old links into `YOUR.PARSE-SERVER.HERE` and `/parse/` respectively. This also adjusts the initial wording for the [REST Quick Reference](https://github.com/montymxb/docs/blob/506928730bdb98949115705bdeb3bb4c64fd1c84/_includes/rest/quick-reference.md) section to remind the user that in the examples provided they should add their own domain, their own mount path and to consider whether they should be accessing over HTTP/HTTPS.

### DIFF
--- a/_includes/android/push-notifications.md
+++ b/_includes/android/push-notifications.md
@@ -480,7 +480,7 @@ curl -X GET \
 -G \
 --data-urlencode 'limit=1000' \
 --data-urlencode 'where={ "city": "San Francisco", "deviceType": { "$in": [ "ios", "android", "winphone", "embedded" ] } }' \
-https://api.parse.com/1/installations
+https://YOUR.PARSE-SERVER.HERE/parse/installations
 ```
 
 If you type the above into a console, you should be able to see the first 1,000 objects that match your query. Note that constraints are always ANDed, so if you want to further reduce the search scope, you can add a constraint that matches the specific installation for your device:
@@ -493,7 +493,7 @@ curl -X GET \
 -G \
 --data-urlencode 'limit=1' \
 --data-urlencode 'where={ “objectId”: {YOUR_INSTALLATION_OBJECT_ID}, "city": "San Francisco", "deviceType": { "$in": [ "ios", "android", "winphone", "embedded" ] } }' \
-https://api.parse.com/1/installations
+https://YOUR.PARSE-SERVER.HERE/parse/installations
 ```
 
 If the above query returns no results, it is likely that your installation does not meet the targeting criteria for your campaign.
--- a/_includes/arduino/other.md
+++ b/_includes/arduino/other.md
@@ -5,7 +5,7 @@ Because the Arduino SDK was designed to minimize memory footprint, it doesn't pr
 For example, you could sign up a user from Arduino through a REST call:
 
 ```cpp
-ParseResponse response = Parse.sendRequest("POST", "/1/users", "{\"username\":\"cooldude6\",\"password\":\"p_n7!-e8\"}", "");
+ParseResponse response = Parse.sendRequest("POST", "/parse/users", "{\"username\":\"cooldude6\",\"password\":\"p_n7!-e8\"}", "");
 ```
 
 In this case, the response will contain the objectId of the created user, assuming it was created successfully.
--- a/_app/main.js
+++ b/_app/main.js
@@ -650,6 +650,9 @@ App.Views = {};
 			// deal with common-lang-blocks
 			this.toggleCommonLangBlocks();
 
+      // setup the server/mount path editor
+      this.setupServerFieldCustomization();
+
 			// add toggles to code blocks if necessary
 			if (this.platform === "ios" || this.platform === "osx" || this.platform === "macos") {
 				new App.Views.Docs.Toggle({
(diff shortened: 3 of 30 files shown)


flovilmart:
Would it be insane to make it dynamic?

montymxb:
Umm possibly. Do you mean as in they would enter their own domain/mount path and all of the examples would populate as such?

montymxb:
@flovilmart looking back at it there's a few hundred references roughly, primarily in the REST guide but also dotted in the other sdk guides as well. Adding in a feature to set their own server url & mount path would require tracking all of those, in addition to setting up the JS to update and editable fields to track those updates.

It could be done if there's an interest in it. But I'll have to go back and set something up.

montymxb:
Added in the ability to set your own server/mount shown on the page. For the other guides they just contain a handful of references, and the server settings takes up space on the side nav bar, so I'm just leaving it out from everything besides the **REST** guide. Makes sense to have it there anyways since all the refs there were changed primarily.

This also makes a quick change to some css that wasn't quite right from #457. Just adds a `float: left` to account for some elements which were breaking down a bit more than they should. Case in point the getting started in the [ios guide](http://docs.parseplatform.org/ios/guide/#getting-started), the `+` should be level rather than breaking down.

 Additionally I moved a change for platform extraction from #443  into **_app/main.js**. I missed it the first time around but that change was made in **assets/js/bundle.js**, which is built up from the former file. If the site were to be rebuilt before it was republished those changes would be erased.

@flovilmart can you take a look at this and give me your thoughts on it? It's tinkering with the sidebar, so I want to be sure this is acceptable before we move this along.

_includes/android/push-notifications.md
```diff
@@ -480,7 +480,7 @@ curl -X GET \
 -G \
 --data-urlencode 'limit=1000' \
 --data-urlencode 'where={ "city": "San Francisco", "deviceType": { "$in": [ "ios", "android", "winphone", "embedded" ] } }' \
-https://api.parse.com/1/installations
+https://YOUR.PARSE-SERVER.HERE/parse/installations
```
flovilmart, line 483: Here?

_includes/android/push-notifications.md
```diff
@@ -493,7 +493,7 @@ curl -X GET \
 -G \
 --data-urlencode 'limit=1' \
 --data-urlencode 'where={ “objectId”: {YOUR_INSTALLATION_OBJECT_ID}, "city": "San Francisco", "deviceType": { "$in": [ "ios", "android", "winphone", "embedded" ] } }' \
-https://api.parse.com/1/installations
+https://YOUR.PARSE-SERVER.HERE/parse/installations
```
flovilmart, line 496: Here?

_includes/arduino/other.md
```diff
@@ -5,7 +5,7 @@ Because the Arduino SDK was designed to minimize memory footprint, it doesn't pr
 For example, you could sign up a user from Arduino through a REST call:
 
 ```cpp
-ParseResponse response = Parse.sendRequest("POST", "/1/users", "{\"username\":\"cooldude6\",\"password\":\"p_n7!-e8\"}", "");
+ParseResponse response = Parse.sendRequest("POST", "/parse/users", "{\"username\":\"cooldude6\",\"password\":\"p_n7!-e8\"}", "");
```
flovilmart, line 8: Here?

flovilmart (changes requested):
A few missing spots but I believe this should work on paper! That’s wonderful otherwise :)

I’ll try it today :)

montymxb:
Ahh, find and replace wasn't nearly as good as I thought it was...one moment, I'll tidy those extra ones up.

montymxb:
Thought I initially missed some find and replaces or something, but yeah those spots have been left out of the 'switching' intentionally. In the other guides there was only sparing use of `api.parse.com` and using `/1/`. Adding in the custom server/mount functionality there would be taking up that extra navbar space for only changing a handful of lines in an entire guide :/. Not sure how assistive it would be if the user would have to hunt just to find the few references it changes.

I'm figuring with it just on the REST guide that's about 90% of our api references using a url & mount, everything else is just a sprinkling of curl refs and the one mention per guide for initializing that sdk (usually localhost). For the extra space it takes up on the left it would seem to have the most direct benefit there.

Just for reference the current platform limiting check is pretty simple, if we do wish to push this onto the other guides as well:
```js
if(this.platform !== 'rest') {
    // only customizes for rest currently
    return;
}
```

montymxb:
This just occurred to me, should we add a toggle for protocol, i.e. a drop down for **http**/**https** to further adjust the relevant examples? In quite a few cases where someone is learning it's likely they may just be testing on localhost.

montymxb:
Okay this is a bit of a jump, but perhaps we can get this on all pages, and in a less intrusive fashion, by making it an expanding section of sorts. If sidebar space is of no consequence then maybe we look into fleshing this out to fully configure with that http/https toggling. 

It would be even _cooler_ to have additional substitution for keys too, like **appId**, **masterKey**, etc. Normally this would be ok as the JS is client side, so no information would be sent across the wire. However the docs are not served securely, so there's no guaranteeing the content on the page hasn't been modified in a malicious fashion, i.e. malicious JS injected by a MITM scraping inserted keys or something like that :/. Still it's food for thought, if we ever got a cert somehow via letsencrypt or something we could get a little bit fancier with this.

flovilmart:
If we put the selection in a reusable section in the header that could work as well? AppIiD and client key would be ok to inject as public, perhaps we could store those in the local data store for reuse. I like that it’s JS dépendant and still renders OK without JS

montymxb:
We can do that. I also realized that the mobile version for smaller screens won't have the sidebar as it becomes a drop down instead, which is just a `<select>` element. To make sure this shows up on mobile (and is reusable) I went ahead and modified it to be a separate section under `common`, which can be added right before the `Get Started` section. A bit cleaner and more manageable that way, and we can add as many customizable fields as we like.

I'll see what I can do about those public key fields, now that this is moving into it's own 'customization' section we can expand it however we like without filling up the sidebar 👍 .

And yeah I can toss those into the local storage to persist across page loads, that would be much more convenient.

montymxb:
@flovilmart are we still interested in pursuing this? If we don't want the hassle for setting up fancy server value substitution we should at least replace all the old refs.

flovilmart:
Sorry, I need to jump back on that!

montymxb:
No rush, just should make a note to get back to this at some point.

flovilmart:
It's beautiful!

flovilmart (approved):

montymxb:
@flovilmart just wanted to say now that this in now we can see what people think of the fields customization in the [REST Guide](http://docs.parseplatform.org/rest/guide/). The other /1/ marks are changed, but just the REST guide has the new customization feature. I imagine if it's working okay I can add a PR later on to add it to other guides, if anyone calls for it.

flovilmart:
that's good for the rest, perhaps we need a reset, button, I had trouble reverting to the original values

montymxb:
Oh, that would be nice. I'll make a note to put up a PR for a reset button.

flovilmart:
Haha thanks!

PR #459, opened 2017-08-06
Fix/update sample code on "Saving Objects" section from iOS Guide documentation

RicardoBarroso:
Fixed Swift syntax on sample code for PFObject "saveInBackground" calls.


flovilmart (approved):

flovilmart:
Thanks!

PR #460, opened 2017-08-06
Update cloud-code.md now that viewing Jobs is supported in parse-dashboard

oallouch:
Just a small edit to tell our users viewing Jobs is supported in parse-dashboard.


flovilmart:
Thanks!

PR #464, opened 2017-08-09
Remove query skip/limit/count constraints documentation

### LINKED ISSUE #461: Documentation regarding query limit and skip limit
For the documentation for each language, there is notes indicating that the maximum query limit is 1000, and the skip limit is 10000. This is outdated information since parse server has no limit on this.

I will link a few of the docs that have this:
[http://docs.parseplatform.org/js/guide/#query-constraints](http://docs.parseplatform.org/js/guide/#query-constraints)
[http://docs.parseplatform.org/dotnet/guide/#query-constraints](http://docs.parseplatform.org/dotnet/guide/#query-constraints)

> Note that the default limit of 100 and maximum limit of 1000 apply to the inner query as well, so with large data sets you may need to construct queries

> You can limit the number of query results returned. The limit is 100 by default but anything from 1 to 1000 is a valid limit:

Here are parse server issues where people have asked the same question and conflicting answers were given:
parse-community/parse-server#25
parse-community/parse-server#917
parse-community/parse-server#548

For confirmation, I have actually created a test script in Javascript that queries items greater than these two limits. They all returned fine.

addisonElliott:
In the old Parse hosted backend, there was a maximum query limit of 1,000 objects, skip limit of 10,000, a maximum rate limit of 160 requests / minute, and a count limit of 1,000 objects.

The Parse Server removed all of those constraints but the documentation was never updated to reflect that. Remove all occurrences of these limits in the documentation

See:
    #461 
    parse-community/parse-server#25
    parse-community/parse-server#917
    parse-community/parse-server#548
    parse-community/parse-server#4088

### DIFF
--- a/_includes/android/queries.md
+++ b/_includes/android/queries.md
@@ -39,7 +39,7 @@ query.whereNotEqualTo("playerName", "Michael Yabuti");
 query.whereGreaterThan("playerAge", 18);
 ```
 
-You can limit the number of results with `setLimit`. By default, results are limited to 100, but anything from 1 to 1000 is a valid limit:
+You can limit the number of results with `setLimit`. By default, results are limited to 100. In the old Parse hosted backend, the maximum limit was 1,000, but Parse Server removed that constraint:
 
 ```java
 query.setLimit(10); // limit to at most 10 results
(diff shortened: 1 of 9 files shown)


flovilmart (approved):

flovilmart:
Thanks!

PR #465, opened 2017-08-09
Add links to download page for Ruby and npm

addisonElliott:
Add links on the README page to the download page for Ruby and npm. Only Bundle has a link and it is likely worthwhile to link to the other two remaining libraries required for installation.


flovilmart:
Thanks!

PR #470, opened 2017-09-01
Documentation about localizing push data.

flovilmart:


acinader:
@flovilmart looks good to me.  Ok to merge?

PR #472, opened 2017-09-16
Change score

ComputerCarl:
Having the player's score match the server port number just causes confusion.


flovilmart:
Haha! yeah :) Thanks!

PR #474, opened 2017-10-11
PushServceApi26 gets added to the manifest automatically now

rogerhu:
@natario1


rogerhu:
Yeah never mind.   This won't work.

PR #481, opened 2017-10-26
Adding documentation about read-only masterKey

flovilmart:


montymxb (approved):
Cool 👍 . ~I didn't check but in the updated README over on the server I'm guessing there's an example of using this?~ I checked, there is, we're good!

PR #488, opened 2017-11-30
Add dot notation key in query examples

acinader:
see https://github.com/parse-community/parse-server/pull/4399


montymxb (dismissed):
Additions look good. We probably want an example for the ios sdk while we're at it though.

acinader:
@montymxb can you take another look at this

TomWFox:
I've fixed a couple of formatting issues. It makes sense to get this merged in even without the iOS example and then we can open an issue.

TomWFox:
@acinader are you happy for me to merge this in?

acinader:
yeah, getting my doc improvement from november of 17 merged in would make me happy.  thanks for asking ;)

TomWFox:
We're really prompt here ;)

TomWFox:
Also I noticed that only the PHP docs have section headings for the different types of query constraints which is much easier to navigate than the other docs - I think I'll open an issue/pr for this too.

PR #495, opened 2017-12-29
Fixes broken mobile navigation

flovilmart:
@montymxb if you check currently the docs on a mobile device, you'll notice the select at the top is broken. This fixes the issue.


montymxb (approved):
Just took a look, can confirm this is the case for the mobile nav. It can be replicated from any browser provided the dimensions are small enough to trigger the mobile layout css. Surprised I haven't noticed this before, nice catch 👍 .

PR #503, opened 2018-02-14
Fix cloud-code averageStars example

sophiehi:
Fixed averageStars example code
Related with #448


flovilmart (approved):

flovilmart:
Thanks for the PR!

PR #504, opened 2018-02-14
Updated getting-started.md

Unic-1:
Added step to set INTERNET permissions which is required for the app to access the server


flovilmart (approved):

flovilmart:
Thanks @Unic-1 !

PR #512, opened 2018-04-04
Renamed incorrect "Option" to "Step"

neelkamath:
It was confusing. But now it isn't.


addisonElliott:
Looks good to me. Haven't looked at the docs in awhile, so I missed that.

PR #514, opened 2018-04-11
Fix typos and warnings

neelkamath:


addisonElliott:
Good catch. I found another instance where `bool` is used instead of `boolean`. It's a few lines above where it says:

> So far we’ve used values with type String, Integer, bool, and ParseObject. Parse also supports float, java.util.Date, and JSONObject.NULL.

Do you mind changing that as well to get them all in one PR?

neelkamath:
I put that single commit of 'renaming "option" to "step"' from the other PR in this one because that time I was too lazy to commit on the same branch and now it'd also cause merge issues.

flovilmart (approved):

flovilmart:
Thanks @neelkamath for the PR and @addisonElliott for the review!

PR #515, opened 2018-04-12
Change Android API Reference Link

addisonElliott:
The current Android API reference is outdated and old. Rather than continuously keep updating it, the JavaDocs are uploaded to Maven and JavaDocs.io will host the documentation for us. It is as simple as changing the link to point and it will stay up to date.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -124,4 +124,4 @@ DEPENDENCIES
   jekyll-redirect-from (= 0.10.0)
 
 BUNDLED WITH
-   1.12.4
+   1.16.1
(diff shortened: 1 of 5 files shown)


flovilmart:
@addisonElliott why don't we do what I've done for iOS and Javascript, namely generate correcly the docs and upload them on each merge to master

https://github.com/parse-community/Parse-SDK-JS/blob/master/.travis.yml#L33
https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/.travis.yml#L33

It's very well integrated with travis and everything should already be setup for that.

PR #519, opened 2018-04-19
Fix #518

### LINKED ISSUE #518: Link header bug
Take [these Android docs](http://docs.parseplatform.org/android/guide/#parcelable) for example, when you click on the `Parcelable` subsection of `Files` it goes all the way up to the `Parcelable` subsection of `Objects` because they have the same name and `Objects` is higher up on the page. Either the pages formatting the docs should be changed or the name of the subsections if they clash.

neelkamath:
Pay special attention to [this commit](https://github.com/parse-community/docs/commit/7c8c59e4513c930dbea269df08b6c2e0c53913d0).


flovilmart:
That seems reasonable. Is that all that'S duplicated?

neelkamath:
Yes, I'm 99% sure. I clicked every header link at least once.

flovilmart:
Very nice! Thanks!

PR #528, opened 2018-06-12
javaScriptKey is required to connect app

thobhanifreddy:
I have to provide javascriptKey to connect my react native app with Parse Server. It took me like an hour to figure it out because it was not mentioned in the docs.


flovilmart:
It’s Only required if you specify one on the server, can you reflect it in the patch ?

thobhanifreddy (commented):
comment added about when javascriptkey is required

flovilmart (approved):
Thanks!

PR #534, opened 2018-07-06
added config example for using with master key.

otmezger:


_includes/js/getting-started.md
```diff
@@ -46,6 +46,16 @@ Parse.initialize("YOUR_APP_ID", "YOUR_JAVASCRIPT_KEY");
 Parse.serverURL = 'http://YOUR_PARSE_SERVER:1337/parse'
 ```
 
+If the Masterkey needs to be provided, use this
```
flovilmart: can you note that the master should only be used in safe environments and never be used on the clientside?

flovilmart (changes requested):

_includes/js/getting-started.md
otmezger: sure, will try to add it.

flovilmart: Thanks a lot, each time we talk about the master key, we should have a big warning ⚠️ or ‼️ To make sure everyone pays attention :)

otmezger:
⚠️ ok, I made the changes and even put the emojis in place :-)

flovilmart:
That’s awesome thanks! Would be awesome to reuse that snippet everywhere :p

PR #535, opened 2018-07-12
Code sample on how to copy a url to Parse.File

acinader:
It took me awhile to figure this out, so wanted to put it somewhere so I can find it again when needed.


flovilmart:
Looks good, which makes me think we could add a helper in the JS SDK to do the download/upload dance

PR #536, opened 2018-07-15
Request to add link to react-parse

doronnahum:
I would appreciate it if you could add a link to my npm package react-parse, https://www.npmjs.com/package/react-parse


flovilmart:
Can you open a pull request just with the necessary changes? I believe you commited the wrong files :)

PR #539, opened 2018-07-23
Update dependency links, remove out of date documentation

Jawnnypoo:
There were a few things here that were outdated or no longer applicable.

Switching over to JitPack, there are some updates to the dependencies that we needed to modify here. This will allow us to move over to one repo more easily eventually.


Jawnnypoo:
@flovilmart Can you take a look at this when you get a chance? Thanks!

_includes/android/getting-started.md
```diff
@@ -1,19 +1,29 @@
...
+	}
+}
+```
 
-```groovy
+Then, add the library to your project `build.gradle`
+```gradle
 dependencies {
-  implementation 'com.parse:parse-android:latest.version.here'
+    implementation "com.github.parse-community.Parse-SDK-Android:parse:latest.version.here"
```
flovilmart, line 23: We renamed all packages? Was there a reason?

_includes/android/user-interface.md
```diff
@@ -177,144 +177,3 @@ You can override any layout resources by having files with the same name as thos
 *   `com_parse_ui_parse_login_help_fragment.xml`: You can change the message for password reset.
 
 We've provided another sample app, [ParseLoginSampleLayoutOverride](https://github.com/parse-community/ParseUI-Android) showing how to do this.  This sample app only has a Facebook login button in `com_parse_ui_parse_login_fragment.xml`, and adds a background image to the login screens.
-
-## ParseQueryAdapter
```
flovilmart: Was this removed completely? What was the rationale?

_includes/android/users.md
```diff
@@ -522,18 +522,3 @@ ParseTwitterUtils.unlinkInBackground(user, new SaveCallback() {
   }
 });
 ```
-
-### Twitter API Calls
```
flovilmart: What that removed as well? Why remove all those as user may rely on them in their implementations?

_includes/android/getting-started.md
Jawnnypoo: Yep! It has to do with repackaging with JitPack instead of deploying to Bintray. This way, it will be a simple move to 1 repo instead of many, and makes deploying new versions easier, as all you need to do is tag a release in GitHub.

_includes/android/user-interface.md
Jawnnypoo: `ListView` in Android is all but deprecated, and I don't think we want to encourage anyone to use it. `ParseQueryAdapter` was built only to be used with `ListView` and not `RecyclerView`. Hopefully over time, we could potentially rebuild `ParseQueryAdapter` to work with `RecyclerView`, but for now I think it is best to remove it to not encourage bad practices.

flovilmart: So mark it as such, deprecate it please. And mark the transition in stress of removing the docs as some users may rely on it! I’d be unhappy to find a removed documentation on any ptoject without a clear explanation or transition plan. Please revert.

_includes/android/users.md
Jawnnypoo: The `ParseTwitterUtils` implementation was using the Apache HTTP client, which was removed from the Android SDK starting with Android 23. The method to sign requests was specifically for the Apache HTTP client. Users could potentially still sign their requests using the `Twitter.getAuthToken()` method (setting the needed headers on the request)

flovilmart: Again, let’s not remove docs, instead, indicate why it was removed and what solution we provide, if any, and starting which version please.

Jawnnypoo:
@flovilmart Okay, adjustments have been made

_includes/android/user-interface.md
```diff
@@ -178,143 +178,145 @@ You can override any layout resources by having files with the same name as thos
...
-Pagination ensures that the table only gets one page of objects at a time. You can set the number of objects are in a page by calling `setObjectsPerPage(int)`.
-
-The query is automatically altered to apply pagination, and a pagination row appears at the bottom of the `AdapterView` to allow users to load the next page.
-
-Pagination is turned on by default. To turn it off, call `setPaginationEnabled(false)`. With pagination turned off, the `ParseQueryAdapter` will use the default `ParseQuery` limit of 100 items.
-
-## Auto-loading of Data
-
-When the `AdapterView` that your `ParseQueryAdapter` is set on is attached to a window, the `ParseQueryAdapter`'s `loadObjects()` method is automatically called, triggering the fetching of the first page of results. To disable this behavior (perhaps to delay the fetching of data, or run some custom logic ahead of time), just call `setAutoload(false)` and call `loadObjects()` manually if autoload is disabled.
+## ParseQueryAdapter (Deprecated 0.0.4)
```
flovilmart, line 181: This is odd that all the lines are changed.. What machine / OS are you on?

Jawnnypoo: Hmm, that is odd. macOS High Sierra.

Jawnnypoo: There were a few lines added, like the deprecation notice. But yes, the rest did stay the same as before.

Jawnnypoo:
@flovilmart Just a quick ping on this, does everything seem good now?

flovilmart (approved):
LGTM

Jawnnypoo:
Awesome. Just FYI, I cannot merge this 😁

flovilmart:
All done!

PR #540, opened 2018-07-24
added "options" parameter to the signature of validateAppId

zsmuller:
In the current Parse-Server guide the description of "Custom authentication" contains this:
> On this module, you need to implement and export those two functions validateAuthData(authData, options) {} and validateAppId(appIds, authData) {}.

However the validateAppId() function is handed a third "options" parameter as well (just as validateAuthData()), so the correct sentence sounds like this:
> On this module, you need to implement and export those two functions validateAuthData(authData, options) {} and validateAppId(appIds, authData, options) {}.

This PR contains only this small change.


flovilmart:
Thanks!

PR #428, opened 2017-04-19
Parcelable documentation for Android

natario1:
We recently added Parcelable support for Android ,

parse-community/Parse-SDK-Android#619
parse-community/Parse-SDK-Android#624
parse-community/Parse-SDK-Android#625

PR #541, opened 2018-07-26
Added requirement to use ignoreAcls() when retrieving data from datastore

jamesncl:
Can affect ParseObjects configured with Role based ACLs. Related to [issue #428 in Android SDK](https://github.com/parse-community/Parse-SDK-Android/issues/428). This note is present in iOS documentation only, but also affects Android. Note that applying query.ignoreAcls() then makes the query unusable for network queries (generates exception "ignoreAcls cannot be mixed with network queries").

### DIFF
--- a/_includes/android/local-datastore.md
+++ b/_includes/android/local-datastore.md
@@ -45,11 +45,23 @@ ParseObject.pinAllInBackground(listOfObjects);
 
 ## Retrieving
 
-Storing objects is great, but it's only useful if you can then get the objects back out later. Retrieving an object from the local datastore works just like retrieving one over the network. The only difference is calling the `fromLocalDatastore` method to tell the `ParseQuery` where to look for its results.
+Storing objects is great, but it's only useful if you can then get the objects back out later. To retrieve an object from the local datastore, call the `fromLocalDatastore` method to tell the `ParseQuery` where to look for its results.
 
 ```java
-ParseQuery<ParseObject> query = ParseQuery.getQuery(“GameScore");
+ParseQuery<ParseObject> query = ParseQuery.getQuery("GameScore");
 query.fromLocalDatastore();
+```
+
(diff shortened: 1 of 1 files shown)


flovilmart (approved):
Perfect! Thanks for the PR!

jamesncl:
I've discovered that calling ignoreAcls() on a query then makes that query unusable for network calls, raising this exception:

`java.lang.IllegalStateException: `ignoreACLs` cannot be combined with network queries`

This is problematic if you want to use a query to load from datastore, then later use the same query to load from network. I can't see any method to "undo ignoreAcls", so you would have to reconstruct the query (ugh). I think a note about this should be added, the problem is I don't exactly know when you really NEED to call ignoreAcls for a data store query so unsure what the advice should be. It seemed to work fine for some of my classes, and not others, and I can't figure out what difference in Roles is responsible (they are all the same as far as I can determine). So now I'm not sure what to do... do we need to revert this change? Or perhaps the documentation amendment needs to be something along the lines of "if you experience problems loading results from the local data store for classes utilising role based ACLs you MAY need to call applyAcls..."?

flovilmart:
You should probably use a factory method that generate you that query, and then run that query on the localDatastore and network independentlyx

jamesncl:
Yeah, this is what I've just implemented in my project. It's quite ugly though, so now think that ignoreAcls should only be used if you need to. Should I create a new pull request to make this clearer? PS also found this on SO, confirms the same issue and suggests an alternative solution (setting ACLs when pinning):
[https://stackoverflow.com/questions/24659129/role-acl-doesnt-work-using-parses-android-local-datastore/31950750#31950750](https://stackoverflow.com/questions/24659129/role-acl-doesnt-work-using-parses-android-local-datastore/31950750#31950750)

flovilmart:
Yep go ahead! Anything that can make it clearer is always welcome!

jamesncl:
Made another pull request, it said it couldn't automatically merge... is that okay? I've got a bit of Git learning to do...

flovilmart:
@jamesncl it's all good. The PR said it couldn't merge because the branch used for the PR was out of sync with the target branch. I just had to fix the conflicts and it's all good!

Thanks for the PR's! It's well appreciated!

PR #542, opened 2018-07-26
Clarify use of ignoreAcls

jamesncl:
Suggesting that you **may** (instead of will) need to use ignoreAcls, only if data is protected by Role based ACL. Also you cannot then use the same query on network (this raises an exception).


flovilmart:
It's all good! Thanks!

PR #543, opened 2018-07-27
add digital ocean spaces config section

tmocellin:
Add a section about the Digital Ocean Spaces configuration with the S3Adapters. 
[#4886](https://github.com/parse-community/parse-server/issues/4886#issuecomment-407028330)


flovilmart:
That’s perfect thanks!

PR #550, opened 2018-08-09
Typo fix & adding periods.

laosb:
I was really thinking that there is some platform called Facobook, and Parse guys are really amazing to officially support such a platform :)


flovilmart (approved):
Haha thanks!

PR #555, opened 2018-09-10
docs: Add JS doc for `Parse.Object.revert()`

NicksonYap:


NicksonYap:
https://github.com/parse-community/Parse-SDK-JS/issues/323

flovilmart:
@NicksonYap thanks for the PR. Does it really fit into the performance section?

NicksonYap:
@flovilmart  you're welcome.

I'm not certain
I think it doesn't but I suppose it can be considered as Client-side caching?
Don't know what name to call this behaviour
To be honest I don't know what's the use of replacing server-side data with client-side data
Since calling `find()` or `get()` the object will trigger an API call anyways

flovilmart:
This isn’t caching. This feature is simply about discarding any changed keys from the client. The use of replacing server side data, is basically the core feature of parse, which is saving and updating objects

TomWFox:
I wonder whether this would make sense in the data section.

Also could we get code samples for the other client SDKs as well?

mtrezza:
I'l go ahead and merge this; could be helpful; fixed husky along the way...

NicksonYap:
Okie, I haven't been in the project for so long i don't recall what this is, all the best!

mtrezza (approved):

PR #558, opened 2018-10-04
Add JS Doc for Local Datastore

dplewis:
https://github.com/parse-community/Parse-SDK-JS/pull/612


_includes/js/local-datastore.md
```diff
@@ -0,0 +1,130 @@
+# Local Datastore
+
+The Parse iOS/OS X SDK provides a local datastore which can be used to store and retrieve `Parse.Object`s. To enable this functionality, call `Parse.enableLocalDatastore()`.
```
acinader: this should be js sdk?

dplewis: Yep

GoMino:
Hello, @dplewis does updating an object while being offline, get sync automatically when network access reappear with this new feature?

Can't wait to test the localstorage feature with the upcoming 2.2.0 release 🤩

_includes/js/local-datastore.md
dplewis: Can you review this again?

_includes/js/local-datastore.md
```diff
@@ -0,0 +1,130 @@
...
+
+There's also a method to remove all objects from default pin
+
+```javascript
+await Parse.Object.unPinAllObjects();
+```
+
+## Pinning with Labels
+
+Manually pinning and unpinning each object individual is a bit like using `malloc` and `free`. It is a very powerful tool, but it can be difficult to manage what objects get stored in complex scenarios. For example, imagine you are making a game with separate high score lists for global high scores and your friends' high scores. If one of your friends happens to have a globally high score, you need to make sure you don't unpin them completely when you remove them from one of the cached queries. To make these scenarios easier, you can also pin with a label. Labels indicate a group of objects that should be stored together.
```
acinader: 1. `individual` should be `individually`
2. the example of 'malloc' and 'free' should be removed.  only a very very small percentage of readers will even know what that is anymore ;).  its not adding value here either.
3. I don't follow your example.  Can you try to re-write it to make the issue more clear (again, I'm not sure what the issue you're trying to point out is)

_includes/js/local-datastore.md
```diff
@@ -0,0 +1,130 @@
...
+await Parse.Object.unPinAllWithName('MyScores', listOfObjects);
+```
+
+There's also a method to remove all objects from a label.
+
+```javascript
+await Parse.Object.unPinAllObjectsWithName('MyScores');
+```
+
+Any object will stay in the datastore as long as it is pinned with any label. In other words, if you pin an object with two different labels, and then unpin it with one label, the object will stay in the datastore until you also unpin it with the other label.
```
acinader: Rather than `Any object`, just `Objects`

here's a take at re-writing: 

> An Object will be kept in the datastore as long as it is pinned by at least one label.  So an object pinned with two labels will stay in the datastore if one of the two labels is unpinned.

acinader (changes requested):
looks good.  one grammar nit and some feedback

_includes/js/local-datastore.md
dplewis: I made the changes and added another section

_includes/js/local-datastore.md
```diff
@@ -0,0 +1,146 @@
...
+
+```javascript
+// Add several objects with a label.
+await Parse.Object.pinAllWithName('MyScores', listOfObjects);
+
+// Add another object with the same label.
+await anotherGameScore.pinWithName('MyScores');
+```
+
+To unpin all of the objects with the same label at the same time, you can pass a label to the unpin methods. This saves you from having to manually track which objects are in each group you care about.
```
acinader: remove 'you care about'

acinader (approved):
lgtm!

PR #559, opened 2018-10-09
Enable localDatastore with Parse.initialize()

djruss70:
Set isLocalDatastoreEnabled = true in the ParseClientConfiguration block used by Parse.initialize(), replacing the legacy method Parse.setApplicationId() for Swift.  Objective C initialisation is unchanged.


flovilmart (approved):
Thanks!

PR #560, opened 2018-10-09
Add Parse Client initialisation

djruss70:
Initialise the Parse client by passing a ParseClientConfiguration block to the Parse.initialize() method, which allows the server URL to be specified (unlike the legacy Parse.setApplicationId() method).


flovilmart (approved):
Thanks!

Issue #556, opened 2018-09-20
Cloud Code doc update  


thobhanifreddy:
Since parse-server v3, cloud code no longer uses callbacks. So we need to update docs accordingly. 

But there are many people out there using older version of parse-server. So not sure how should we update it.

flovilmart:
Feel free to get started on this.

acinader:
I'm gonna close this as too general.  I read through the cloud code just now and I see no obviously wrong code examples.

PR #566, opened 2018-10-18
updated afterSave with promises

thobhanifreddy:
As a part of https://github.com/parse-community/docs/issues/556 I tried a updating afterSave with promises. Can someone please review it and suggest any improvements? Looking forward to updated other parts if this PR is upto the point.

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -226,16 +226,25 @@ Parse.Cloud.beforeSave("Review", (request) => {
 In some cases, you may want to perform some action, such as a push, after an object has been saved. You can do this by registering a handler with the `afterSave` method. For example, suppose you want to keep track of the number of comments on a blog post. You can do that by writing a function like this:
 
 ```javascript
-Parse.Cloud.afterSave("Comment", (request) => {
-  const query = new Parse.Query("Post");
-  query.get(request.object.get("post").id)
-    .then(function(post) {
+Parse.Cloud.afterSave("Comment", async request => {
+  try {
+    const query = new Parse.Query("Post");
+    const post = await query.get(request.object.get("post").id);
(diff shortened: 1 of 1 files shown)


flovilmart:
When using afterSave with await, you’ll be blocking the response sent to the client till your async method completes.

Is that what you intended?

thobhanifreddy:
@flovilmart I thought async/await is cleaner than `.then`. Otherwise they work almost the same.

flovilmart:
> I thought async/await is cleaner than .then. Otherwise they work almost the same.

I believe the key word is *almost*. 

consider the following:

```
function myAfterSave0(req) {
    const query = new Parse.Query();
    query.find().then((results) => {
        // do something...
   })
} 
function myAfterSave1(req) {
    const query = new Parse.Query();
    // return the promise
    return query.find().then((results) => {
        // do something...
   })
}

async function myAfterSave2(req) {
    const query = new Parse.Query();
    // await the results
    const results =  await query.find();
    /// do something
} 
```

Now imagine I call those 3 functions and insect the object

```
const res0 = myAfterSave0()
const res1 = myAfterSave1()
const res2 = myAfterSave2()
```

- [ ] `typeof res0 === 'undefined'`
- [ ] `res0 instanceof Promise`
- [ ] `typeof res1 === 'undefined'`
- [ ] `res1 instanceof Promise`
- [ ] `typeof res2 === 'undefined'`
- [ ] `res2 instanceof Promise`

In afterSave parse-server checks the type that is retuned from the handler. if the handler returns a promise, then the server will wait till the promises resolves.

So while it's almost the same it isn't.

acinader:
i think #578 addresses this in a better way, i think we can close this and we'll resolve via working out #578 @thobhanifreddy let me know if you think that we should re-open this instead....

PR #570, opened 2018-11-05
Update push-notifications.md

### LINKED ISSUE #569: 📙 Probably there is no Int() function or class
I think there ambiguity in [silent push notification](https://docs.parseplatform.org/parse-server/guide/#silent-notifications) docs, when its says **Int(1)** , actually it means 1 not "1" and I suspect there is any function/class or library (as per my knowledge) named **Int**.  A better wording could have been **Number(1)** , instead of **Int(1).** Please correct me if I am wrong.

sakhunzai:
Int() is not a common function across all programming languages, this removes the ambiguity.

Fixes #569.

### DIFF
--- a/_includes/parse-server/push-notifications.md
+++ b/_includes/parse-server/push-notifications.md
@@ -276,7 +276,7 @@ If you're interested in any of these features, [don't hesitate to jump in and se
 
 ### Silent Notifications
 
-If you have migrated from Parse.com Push and you are seeing situations where silent notifications are failing to deliver, please ensure that your payload is setting the `content-available` attribute to Int(1) and not "1". This value will be explicitly checked.
+If you have migrated from Parse.com Push and you are seeing situations where silent notifications are failing to deliver, please ensure that your payload is setting the `content-available` attribute to Int(1) (or just 1 as in javascript) and not "1". This value will be explicitly checked.
 
 ### PPNS
 


flovilmart:
Perhaps it would be even better to show the payload as it would remove to complete ambiguity. I’m not too attached to the `Int`.
We likely should say that 1 has to be a valid JSON number.

https://www.w3schools.com/js/js_json_datatypes.asp

sakhunzai:
sure much better, but I am not sure what should be wording :)

flovilmart:
What do you mean by that?

TomWFox:
@sakhunzai could you have a go at wording it? And then we can try and get this merged.

_includes/parse-server/push-notifications.md
```diff
@@ -276,7 +276,7 @@ If you're interested in any of these features, [don't hesitate to jump in and se
 
 ### Silent Notifications
 
-If you have migrated from Parse.com Push and you are seeing situations where silent notifications are failing to deliver, please ensure that your payload is setting the `content-available` attribute to Int(1) and not "1". This value will be explicitly checked.
+If you have migrated from Parse.com Push and you are seeing situations where silent notifications are failing to deliver, please ensure that your payload is setting the `content-available` attribute to Int(1) (or just 1 as in javascript) and not "1". This value will be explicitly checked.
```
TomWFox, line 279: Perhaps something like this:
> please ensure that your payload is setting the `content-available` attribute to the integer 1 and not the string "1".

TomWFox: @parse-community/server does my suggested change sound ok?

acinader:
lgtm!

PR #577, opened 2018-12-10
Improve docs for tvos

toto:
Small update to explicitly state tvOS support and describe it's config in relation to [this PR of the iOS/tvOS/macOS SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1375). Took me a while to figure this out so this might help a few people.


flovilmart (approved):
This looks good! Should we add a word that this will be in the future releases of the Obj-C SDK?

PR #578, opened 2018-12-12
After save details

acinader:
In my first attempt at upgrading to parse 3, I did it in such a way that our afterSave hooks were now blocking the response, so I looked into what was going on and figured it should be better documented.

Try to improve afterSave documentation to include:

1. How to ensure that afterSave runs async if necessary
2. That request.context is a feature and how to use it.

I also added a real-world code example that uses modern javascript.


acinader:
doh.  I didn't notice #566 I'll take a look....

flovilmart (approved):
LGTM!

PR #586, opened 2019-02-23
Update usage.md

BufferUnderflower:
Could not find the reason why i can't upload files (~2 mb) running parse-server on dokku (nginx)


acinader (approved):
thanks @BufferUnderflower

PR #597, opened 2019-03-13
Update code snippets for sign up & log in view controller

TomWFox:
There is one remaining issue that I have noticed being that the link to [PFSignUpView](http://parseplatform.org/Parse-SDK-iOS-OSX/api/Classes/PFSignUpView.html), which is at the end of the PFSignUpViewController section, 404s.


TomWFox:
@parse-community/ios-osx Any ideas where the API Reference for PFSignUpView can be found?

TomWFox:
I'm going to merge this for now and will update the links at some point.

PR #601, opened 2019-03-13
Update push-notifications.md

jasonkneen:
Fixed incorrect JSON that causes "improper use of $inQuery" errors -- now correctly works assuming the _Installation class has a "user" pointer to the _User class.


TomWFox:
Could someone from the @parse-community/server team take a look at this?

acinader:
lgtm

PR #604, opened 2019-03-17
Remove outdated push notification content

TomWFox:
The option of client push has been removed from parse server, thus I have removed all sections that involve sending push notifications from the iOS SDK.

I have also tried to make it as clear as possible that client push is not allowed anymore and updated references with that in mind.

Addresses problems brought up by #405.


TomWFox:
@acinader I know you don’t use the iOS SDK but could you cast an eye over this just to check that the changes seem in line with the removal of the ability of client push from parse server.

dplewis (approved):
LGTM

PR #606, opened 2019-03-21
Example for extending Parse.User

dplewis:
https://github.com/parse-community/Parse-SDK-JS/issues/722

Should we always commit `assets/js/bundle.js`?


acinader (changes requested):
I think this is a good improvement, but we should fix the sdk to improve how it handles the static methods.

Here's how I did it for the php-sdk: https://github.com/parse-community/parse-php-sdk/pull/394/files

and here's an example of how to fix it for the JS-SDK: https://runkit.com/arthur/static-factory-w-subclass

You up for making the changes?

_includes/js/objects.md
```diff
@@ -89,6 +89,36 @@ However, when using `extends`, the SDK is not automatically aware of your subcla
...
+  }
+
+  doSomething() {
+    return 5;
+  }
+}
+Parse.Object.registerSubclass('CustomUser', CustomUser);
+```
+
+In addition to queries, `logIn` and `signUp` returns subclass.
```
acinader: will return the subclass `CustomUser`

dplewis:
Sounds like an easy fix. I didn't know about `new this()` in static functions

_includes/js/objects.md
```diff
@@ -89,6 +89,36 @@ However, when using `extends`, the SDK is not automatically aware of your subcla
 Parse.Object.registerSubclass('Monster', Monster);
 ```
 
+Similarly, you can `extends` `Parse.User`
```
TomWFox: Maybe this should say "Similarly, you can use `extends` with `Parse.User`"

dplewis: Sounds good

dplewis:
@Tom-Fox1 What is `bundle.js` used for?

TomWFox:
@dplewis sorry, I don’t know, I meant to look into it but haven’t yet

TomWFox (approved):

acinader:
@dplewis @Tom-Fox1 

I just took a look and here's what I learned...

From looking at the `package.json`, I'm inferring the following (with high confidence):

1. since we're deploying to Github pages, the site needs to be static.
1. to make development sane, it is good to break the docs down into a set of files that help with headers, navigation, implying links, etc.
1. `bundle.js` is the 'built' javascript file that has all of the sources 'compiled' into it so the whole doc set can be served from that one static file that requires no 'runtime' when a page is requested.
2. during development, we want to watch changed files and dynamically serve up the current state of the files as they are edited.  This is done with `npm run dev` which runs the `dev` script in `package.json` which uses jekyll to build the html & webpack to build the js 'on the fly'.
3. after you've made all your changes, you'll want to commit to git so you can then push and create a PR on github.  In package.json, we use [husky](https://github.com/typicode/husky#readme) to catch the commit and first call the npm script `webpack` which tells webpack to create a production build of `bundle.js` from the state of the files and then add it to the commit.
4. we don't need to anything about the html because git hub pages supports jekyll and it'll look for [_config.yml](https://github.com/parse-community/docs/blob/gh-pages/_config.yml) and then use that to build the html.

So the short answer is: `bundle.js` can be ignored by us in pr's and will automagically be changed any time there's a change.

and there's a TODO for @Tom-Fox1: take a look at _config.yml cause it has our tags for severfault (whatever that is) and stackoverlow.

now i understand the doc stuff much better....:).

dplewis:
@acinader Thanks for the detailed description. I understand it much better.

PR #607, opened 2019-03-22
Add language toggles to REST docs

TomWFox:
Some sections are missing the `<div class="language-toggle">` and `</div>` so the code displays but not with the toggle and as two separate code blocks instead.

Sections to do / check:
- [x] Getting Started
- [x] Quick Reference
- [x] Objects
- [x] Queries - did discover some other formatting issues #608
- [x] Users
- [x] Sessions (changes made)
- [x] Roles
- [x] Files
- [x] GeoPoints
- [x] Data
- [x] Push Notifications (changes made)
- [x] Config (changes made)
- [x] Analytics
- [x] Cloud Code
- [x] Schema
- [x] Hooks (changes made)
- [x] Security
- [x] Error Codes

### DIFF
--- a/_includes/rest/config.md
+++ b/_includes/rest/config.md
@@ -39,7 +39,6 @@ The response body is a JSON object containing all the configuration parameters i
 
 You can also update the config by sending a `PUT` request to config URL. Here is a simple example that will update the `Parse.Config`:
 
-<div class="language-toggle">
 <pre><code class="bash">
 curl -X PUT \
   -H "X-Parse-Application-Id: <span class="custom-parse-server-appid">${APPLICATION_ID}</span>" \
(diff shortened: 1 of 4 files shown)


TomWFox:
@acinader should there be a JavaScript code snippet here? - [config section](https://docs.parseplatform.org/rest/guide/#config)

Screenshot 2019-03-22 at 20 04 58

acinader:
oh, I trust your judgment in this regard more than mine.

I'm not worried about the consistency, but the experience with the python button if you click on it is bad.

TomWFox:
I think I will just remove the language toggle - there's no way I can write a python example - I've only used python for a school assessment that was written off because of 'cheating' (not by me) 😂

PR #611, opened 2019-03-29
Add docs for passing masterKey to user._linkWith

acinader:


acinader:
@Tom-Fox1 fyi

_includes/js/users.md
```diff
@@ -474,25 +474,49 @@ The body of the response will contain the `objectId`, `createdAt`, `sessionToken
...
+
 ### Unlinking
 
 Unlinking an existing user with a service also uses \_linkWith() method to clear `authData` from the user by setting the `authData` for the service to `null`.  For example, unlinking a user with a Facebook account would use a request like this:
 
 ```javascript
-let user = new Parse.User();
+const user = new Parse.User();
 user.id = "uMz0YZeAqc";
 user._linkWith("facebook", {}).then(function(user){
```
dplewis: `user._linkWith("facebook", {})` I don't think this works.

should be replaced with `_unlinkFrom` or `user._linkWith("facebook", { authData: null })`

_includes/js/users.md
```diff
@@ -474,25 +474,49 @@ The body of the response will contain the `objectId`, `createdAt`, `sessionToken
...
+  'CustomAdapter',
+  myAuthData,
+  { useMasterKey: true }
+);
+```
+
+On rest, web, mobile, or TV clients, users can then login using the `CustomAdapter` by passing `myAuthData`:
+
+```javascript
+const loggedIn = await Parse.User.logInWith('CustomAdapter', myAuthData);
```
dplewis, line 510: An example of CustomAdapter, server config and `Parse.User._registerAuthenticationProvider`  should be included.

dplewis (changes requested):

TomWFox (approved):
I fixed the link to the 3rd party auth section in the server guide.

I can't comment on the code but the wording looks good to me!

dplewis:
Good catch!

PR #612, opened 2019-04-10
Remove Arduino guide

TomWFox:
As suggested by @flovilmart I have archived the Parse-Arduino-SDK repo as there seems to be a lack of use & contribution. Thus I would think it makes sense to remove the guide from the docs.


dplewis:
I don’t think the docs should be removed. There are users who need it and the archived project maybe revived someday.

TomWFox:
True, I think we should at least add a warning at the start of the guide.

TomWFox:
Although it does seem unlikely that anyone is still using it considering there have been like one or two contributions since Parse.com was shut down and it seems it still is not fully set up for use with Parse Server out of the box.

TomWFox:
@acinader thoughts?

TomWFox:
I will leave this for now as it may be best to totally remove the guide at some point but for now I have made a new PR #616 to add a warning.

PR #615, opened 2019-04-23
Improve linking user documentation

dplewis:


TomWFox (approved):
LGTM - I just improved the capitalisation a bit

acinader:
don't wait on me.  If its documentation and approved by @TomWFox it's good to go!

dplewis:
@acinader I've been looking through comments and posts about linking user. And I confused myself.

From my understanding Parse-Server supports 3rd Party Auth and Custom Auth.

The JS SDK only supports Facebook, Anonymous and Custom Auth. If users want other 3rd Party they would have to create their own provider (i.e Twitter, WeChat etc) and call `Parse.User._registerAuthenticationProvider(provider);` (not documented).

I can guess that the parse.com guys didn't want to maintain those auth providers, or those 3rd Party Auth came from Parse Open Source community and the JS SDK hasn't been updated.

What should we do going forward?

Edit: _linkWith and _unlinkFrom can't do much without `Parse.User._registerAuthenticationProvider(provider);` and should be removed from docs

acinader:
Do you have a suggestion?

Offhand, my thought would be to document how to do it.

I definitely had a hard time figuring out how to set up a custom provider, and you helped me at one point.

dplewis:
I think we can document it the same way users have to register their own subclasses.

dplewis:
Can one of you guys pull this down and see if `Parse.User._registerAuthenticationProvider` documentation is clear?

TomWFox:
@dplewis I've read it over and it seems ok but it's hard for me to say without actually trying it out, which I can't do right now.

_includes/js/users.md
```diff
@@ -313,15 +313,13 @@ When this code is run, the following happens:
 1.  The user is shown the Facebook login dialog.
 2.  The user authenticates via Facebook, and your app receives a callback.
 3.  Our SDK receives the Facebook data and saves it to a `Parse.User`. If it's a new user based on the Facebook ID, then that user is created.
-4.  Your `success` callback is called with the user.
 
-You may optionally provide a comma-delimited string that specifies what [permissions](https://developers.facebook.com/docs/authentication/permissions/) your app requires from the Facebook user.  For example:
+You may optionally provide a comma-delimited string that specifies what [permissions](https://developers.facebook.com/docs/authentication/permissions/) your app requires from the Facebook user. For example:
```
acinader, line 317: ha.

_includes/js/users.md
```diff
@@ -251,53 +251,53 @@ Using our Facebook integration, you can associate an authenticated Facebook user
...
 ```
 
 The function assigned to `fbAsyncInit` is run as soon as the Facebook JavaScript SDK has completed loading. Any code that you want to run after the Facebook JavaScript SDK is loaded should be placed within this function and after the call to `Parse.FacebookUtils.init()`.
 
 If you encounter any issues that are Facebook-related, a good resource is the [official getting started guide from Facebook](https://developers.facebook.com/docs/reference/javascript/).
 
-If you encounter issues that look like they're being returned from Parse's servers, try removing your Facebook application's App Secret from your app's settings page.
-
-There are two main ways to use Facebook with your Parse users: (1) logging in as a Facebook user and creating a `Parse.User`, or (2) linking Facebook to an existing `Parse.User`.
+There are two main ways to use Facebook with your Parse users: (1) [logging in as a Facebook user](#login--signup) and creating a `Parse.User`, or (2) [linking Facebook](#linking) to an existing `Parse.User`.
```
acinader, line 292: A `ParseUser` can be associated with Facebook either at signup or by linking an existing user.

* [Signing up with Facebook](#login--signup)
* [Linking an existing user to Facebook](#linking)

acinader: maybe?

_includes/js/users.md
```diff
@@ -362,68 +362,60 @@ Our library manages the `FB` object for you. The `FB` singleton is synchronized
...
-  }
-}
-```
-
-Learn more about [Twitter login](https://dev.twitter.com/docs/auth/implementing-sign-twitter).
+const authData = {
+  "id": "12345678",
+  "screen_name": "ParseIt",
+  "consumer_key": "SaMpLeId3X7eLjjLgWEw",
+  "consumer_secret": "SaMpLew55QbMR0vTdtOACfPXa5UdO2THX1JrxZ9s3c",
```
acinader: what credentials are these?

acinader (approved):
lgtm!

_includes/js/users.md
dplewis: Fake credentials. This code doesn’t work because twitter docs need to be updated too. Can you provide feedback https://github.com/parse-community/Parse-SDK-JS/issues/771#issuecomment-486735957

acinader:
@dplewis this can be merged, right?

dplewis:
@acinader I forgot about this one. Just a quick update.

PR #620, opened 2019-04-30
Remove unneeded parse.com references

### LINKED ISSUE #619: 📙 Remove unneeded references to Parse.com
__Link to section:__
Dotted throughout all the guides in different sections.

__What is the issue?__
There are still lots of references in the docs to Parse.com or explanations that use language that imply use on the hosted Parse Cloud which no longer exists.

e.g.

The [sending pushes](https://docs.parseplatform.org/js/guide/#sending-pushes) section in the JS guide...
> Sending notifications is often done from the Parse.com push console

As appropriate references should be removed or adjusted to language that is correct for Parse Server.

e.g.
> Sending notifications is often done from the Parse Dashboard push console

instead of...
> Sending notifications is often done from the Parse.com push console

TomWFox:
Fixes #619.


TomWFox:
There are two links that I have had to remove from most of the guides because they references the old Parse.com blog.

http://blog.parse.com/learn/engineering/implementing-scalable-search-on-a-nosql-backend/
http://blog.parse.com/2014/09/03/the-dangerous-world-of-client-push/

Ideally I'd like to replace rather than remove these - the client push post is on the new blog but is out of date and I assume the other post is also on the new blog.

PR #621, opened 2019-05-31
add master key and rest api key for rest doc

nasermirzaei89:


TomWFox (approved):
Seems good to me, just thinking we should add a note at the start of the docs to remind people that their master key should be kept secret. @acinader are you happy with these changes?

PR #622, opened 2019-06-02
Handling of task result in Caching Queries

git-elliot:
In caching queries using local datastore enabled, I've added how to handle the result of a task, by first checking if there is no exception and then converting result to list of ParseObjects.


rogerhu (approved):

TomWFox:
@parasthekoder thanks for the contribution!

PR #626, opened 2019-06-14
Fix broken link in third-party-auth.md

W0lfw00d:
The link to the twitter docs was broken. It's now set to the correct url.


TomWFox (approved):
Thanks for the fix!

PR #629, opened 2019-06-22
Minor spelling fix

rlevy12:


rlevy12:
Is the Webservice protocol wss:// or ws:// ? I've seen it referenced differently in various documents and forums. Thanks!

TomWFox (approved):
Thanks for the fix

TomWFox:
I don't know about the the ws:// etc but I did a DuckDuckGo search and came across [this](https://stackoverflow.com/questions/46557485/difference-between-ws-and-wss)

rlevy12:
Thanks for the link - that was helpful. I was hoping that the ws/wss thing was my issue, but alas it was not. I have a LiveServer setup on the same server as regular working Parse server (both https), but my client (browser) shows the same errors in the console every 15-20 seconds, even when I know no data is being updated: WebSocket connection to 'wss://my-server-name.com: Unexpected response code: 403. There is no load balancer between the client and the Parse server (no nginx etc.). I've Googled this for hours and hours on end and couldn't find a solution. Any ideas?

TomWFox:
Sorry, not sure about this. If you think this is a bug please open an issue on the Parse Server repo with as much detail as you can provide, if not perhaps ask a question on stack overflow or the community forum where appropriate.

PR #635, opened 2019-07-12
Read preference

### LINKED ISSUE #5515: missing docs for setting the readPreference option in a query
Hi,
Like in my previous [issue](https://github.com/parse-community/parse-server/issues/5508) (where I said I can't find user.getSessionToken() in the docs), I face another documentation problem : this time I'd like to make some queries read from the secondary mongodb server, but I can't find the readPreference in the reference js documentation.

Thx a lot in advance,
Olivier

davimacedo:
close https://github.com/parse-community/parse-server/issues/5515


dplewis:
Can you add a section for php? I can do a PR later

davimacedo:
Done for php as well

_includes/rest/objects.md
```diff
@@ -165,6 +165,33 @@ print result
 </code></pre>
 </div>
 
+When using a MongoDB replica set, you can use the `readPreference` option to choose from which replica the object will be retrieved. You can also use the `includeReadPreference` option to choose from which replica the included pointers will be retrieved. The possible values for both options are `PRIMARY` (default), `PRIMARY_PREFERRED`, `SECONDARY`, `SECONDARY_PREFERRED`, or `NEAREST`. If the `includeReadPreference` option is not set, the same replica chosen for `readPreference` will be also used for the includes.
+
```
dplewis, line 169: Same rule applies to subqueryReadPreference

dplewis (changes requested):
Quick nit

_includes/js/queries.md
```diff
@@ -537,3 +537,15 @@ query.distinct("age")
     // There was an error.
   });
 ```
+
+## Read Preference
+
+When using a MongoDB replica set, you can use the `query.readPreference(readPreference, includeReadPreference, subqueryReadPreference)` function to choose from which replica the objects will be retrieved. The `includeReadPreference` chooses from which replica the included pointers will be retrieved and the `subqueryReadPreference` chooses in which replica the subqueries will run. The possible values are `PRIMARY` (default), `PRIMARY_PREFERRED`, `SECONDARY`, `SECONDARY_PREFERRED`, or `NEAREST`. If the `includeReadPreference` is not passed, the same replica chosen for `readPreference` will be also used for the includes. The same rule applies for the `subqueryReadPreference`.
```
TomWFox: Maybe... “The includesReadPreference chooses” would read better as “The includesReadPreference argument/parameter chooses”

TomWFox: Same for the others as well.

TomWFox (approved):
Looks good, just a minor wording suggestion.

_includes/rest/objects.md
davimacedo: When getting an object, the `subqueryReadPreference` does not make sense. I think you can even pass it, but it will never do any thing.

_includes/js/queries.md
davimacedo: done

_includes/rest/objects.md
davimacedo: Do you agree to not mention it?

dplewis: I agree

PR #636, opened 2019-07-12
Adding header to deleting files section for visibility.

noobs2ninjas:
Removed information advising to go to app settings to add file cleanup request as this is not available with Parse Dashboard.


TomWFox:
Thanks for contributing!

PR #638, opened 2019-07-12
Adding header to deleting files section on iOS docs for visibility.

noobs2ninjas:
Removed information advising to go to app settings to add file cleanup request as this is not available with Parse Dashboard.

### DIFF
--- a/_includes/ios/files.md
+++ b/_includes/ios/files.md
@@ -142,6 +142,6 @@ file?.saveInBackground({ (success: Bool, error: Error?) in
 ```
 </div>
 
-You can delete files that are referenced by objects using the [REST API]({{ site.baseUrl }}/rest/guide/#deleting-files). You will need to provide the master key in order to be allowed to delete a file.
+## Deleting Files
 
-If your files are not referenced by any object in your app, it is not possible to delete them through the REST API. You may request a cleanup of unused files in your app's Settings page. Keep in mind that doing so may break functionality which depended on accessing unreferenced files through their URL property. Files that are currently associated with an object will not be affected.
+You can delete files that are referenced by objects using the [REST API]({{ site.baseUrl }}/rest/guide/#deleting-files). You will need to provide the master key in order to be allowed to delete a file. If your files are not referenced by any object in your app, it is not possible to delete them through the REST API. 


TomWFox (approved):
Thanks for contributing! @davimacedo this looks ok to me, would you be able to take a look as well?

_includes/ios/files.md
```diff
@@ -142,6 +142,6 @@ file?.saveInBackground({ (success: Bool, error: Error?) in
 ```
 </div>
 
-You can delete files that are referenced by objects using the [REST API]({{ site.baseUrl }}/rest/guide/#deleting-files). You will need to provide the master key in order to be allowed to delete a file.
+## Deleting Files
 
-If your files are not referenced by any object in your app, it is not possible to delete them through the REST API. You may request a cleanup of unused files in your app's Settings page. Keep in mind that doing so may break functionality which depended on accessing unreferenced files through their URL property. Files that are currently associated with an object will not be affected.
+You can delete files that are referenced by objects using the [REST API]({{ site.baseUrl }}/rest/guide/#deleting-files). You will need to provide the master key in order to be allowed to delete a file. If your files are not referenced by any object in your app, it is not possible to delete them through the REST API.
```
davimacedo, line 147: > If your files are not referenced by any object in your app, it is not possible to delete them through the REST API.

It is actually possible. If you use the REST API, it will delete the file in the storage no matter if it is referenced by an object or not.

davimacedo (changes requested):

_includes/ios/files.md
TomWFox: In that case I think the JS guide is misleading as well.
It currently says...
> You can delete files that are referenced

Surely it should say something like "You can delete files using the REST API" or "You can delete referenced or unreferenced files using the REST API"

noobs2ninjas: I agree 100%. In fact I think we should relay three things here.  1) Deleting an object DOES NOT delete the file when using the mongoDB grid bucket(Not sure about adapters...clarification?). 2) That it's recommended to call the rest API to delete the file prior to an objects deletion. 3) That this is the ONLY method for deleting files referenced or otherwise.  

I was merely trying to draw attention to the section by adding a header and editing out old documentation that is no longer relevant as it very much to the naked eye while skimming seems part of the Progress documentation. However if you guys agree Ill edit further to relay this information.

davimacedo: That would be nice!

Regarding your points:
1) That's true for all adapters.
2) I don't think it is a recommendation; some developers can choose to keep the files there; the same file can be used by different objects and that's why they are not deleted automatically.
3) Yes. The only way to delete the file (referenced or not) is through the REST API.

Additionally, since you're already in the process, in the REST Guide, you will find:
"Note that deleting a PFObject with a file associated with it will not delete the file. All files stored on Parse should be deleted by using the above explained API."

Instead of `a PFObject` (too specific only for iOS), can you please change to `an object`?

TomWFox: I agree with @davimacedo, those changes would be great 👍

noobs2ninjas: Ok Ill get to work on that. Im also going to give the whole thing a look over to see if we have any deprecations that we want to address. 

I think we should at least mention that parse does not provide a way to find unreferenced file names. I think saying that you can delete files via rest whether the file is referenced or not might lead a newer developer to believe that there must also be an easy way to find unreferenced files. Back when I started my career as a programmer I used parse because I didn't know much about databases or file structures. So I leaned heavily on that cleanup button. I would have been clueless as what to do if I had to find unreferenced filenames myself after the fact.

TomWFox:
@noobs2ninjas I think you addressed this in #650, so ok to close?

noobs2ninjas:
I think we are good to go.

PR #641, opened 2019-07-14
Fix language selector

### LINKED ISSUE #643: 🐛 iOS SDK Guide language selection broken for examples
Within the iOS SDK guide example code is provided in Objc and Swift.  Unfortunately only the Objc example code displays.  If you select the Swift language - nothing happens, the example code is not updated.

Checking the browser console I can see the following error message.
Screenshot 2019-07-14 at 21 03 32
Unfortunately my JS skills are <cough /> worse than poor.

I can see that the bundle.js file was updated only 2 days ago - I guess something in that update has caused this error.

**To Reproduce**
1. Go to https://docs.parseplatform.org/ios/guide/#getting-started
2. Scroll down to 'Saving Objects'
3. Click the SWIFT example button

**Expected behavior**
the example code should update to display the Swift example code.

**Screenshots**

 - OS: Mac OS / 10.14/ Mojave
 - Safari, Firefox, Chrome

TomWFox:
As identified here - https://community.parseplatform.org/t/parse-ios-sdk-guide-language-selection-broken/556

The culprit was a jquery update.

Fixes #643


davimacedo (changes requested):
Can you please pin 3.4.1 instead? I've just tested here and it seems that they have fixed the bug in the newest version.

TomWFox:
@davimacedo even better! Done, I tested it as well and it seems good!

TomWFox:
I want to get this in so I’ve gone ahead and merged, should be fine 😉

PR #646, opened 2019-07-24
Remove erroneous docs for Jobs in 3.x

woodardj:
Looks like this line got left behind during the process of updating the docs to 3.0. (jobs are no longer called with a `status` argument.


acinader (approved):
thanks @woodardj

PR #652, opened 2019-07-29
GraphQL Customisation

omairvaiyani:
This is a guide for the [GraphQL Config feature](https://github.com/parse-community/parse-server/pull/5782) but I have placed it under the header "Customisation". I think this would be an appropriate place to also add docs for this [PR](https://github.com/parse-community/parse-server/pull/5821)

### DIFF
--- a/_includes/graphql/customisation.md
+++ b/_includes/graphql/customisation.md
@@ -81,9 +81,9 @@ interface ParseGraphQLConfiguration {
 
 ### Set or Update Configuration
 
-We have provided a public API in ```ParseGraphQLServer``` which accepts the above JSON object for setting (and updating) your Parse GraphQL Configuration, ```setGraphQLConfig```:
+We have provided a public API in `ParseGraphQLServer` which accepts the above JSON object for setting (and updating) your Parse GraphQL Configuration, `setGraphQLConfig`:
 
-```typescript
+```javascript
   const parseGraphQLServer = new ParseGraphQLServer(parseServer, {
     graphQLPath: parseServerConfig.graphQLPath,
     playgroundPath: parseServerConfig.playgroundPath
@@ -98,14 +98,15 @@ We have provided a public API in ```ParseGraphQLServer``` which accepts the abov
 
 ### Include or Exclude Classes
 
-By default, all of your Parse classes, including the defaults such as ```Parse.User```, ```Parse.Session```, ```Parse.Role``` are added to the schema. You can restrict this using the ```enabledForClassess``` or ```disabledForClassess`` options, which accepts an array of Class names.
+By default, all of your Parse classes, including the defaults such as `Parse.User`, `Parse.Session`, `Parse.Role` are added to the schema. You can restrict this using the `enabledForClassess` or `disabledForClassess` options, which accepts an array of class names.
 
-In the following example, we limit our GraphQL schema to only expose the default ```_User``` class, along with a few custom classes:
+
+In the following example, we limit our GraphQL schema to only expose the default `_User` class, along with a few custom classes:
 
 ```javascript
 {
-  enabledForClasses: ["_User", "Book", "Review", "Comment"],
-  disabledForClasses: null
+  "enabledForClasses": ["_User", "Book", "Review", "Comment"],
+  "disabledForClasses": null
 }
 ```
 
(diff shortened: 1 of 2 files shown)


_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+      update?: boolean;
+      destroy?: boolean;
+    };
+  }>
+}
+```
+
+### Set or Update Configuration
+
+We have provided a public API in ```ParseGraphQLServer``` which accepts the above JSON object for setting (and updating) your Parse GraphQL Configuration, ```setGraphQLConfig```:
```
TomWFox: Should be single backticks

_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+  const config = {
+    // ... ParseGraphQLConfiguration
+  };
+
+  await parseGraphQLServer.setGraphQLConfig(config);       
+```
+
+### Include or Exclude Classes
+
+By default, all of your Parse classes, including the defaults such as ```Parse.User```, ```Parse.Session```, ```Parse.Role``` are added to the schema. You can restrict this using the ```enabledForClassess``` or ```disabledForClassess`` options, which accepts an array of Class names.
```
TomWFox:
```suggestion
By default, all of your Parse classes, including the defaults such as ```Parse.User```, ```Parse.Session```, ```Parse.Role``` are added to the schema. You can restrict this using the ```enabledForClassess``` or ```disabledForClassess`` options, which accepts an array of class names.
```

_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+  // undefined or null results in the default behaviour, i.e. include all classes
+  enabledForClasses: undefined,
+  // override the included classes by filtering out the following:
+  disabledForClasses: [ "UserSensitiveData", "ProductOrder", "Invoice" ]
+}
+```
+
+### Input Types
+
+By default, we enrich the schema by generating a number of [Input Types](https://graphql.org/learn/schema/#input-types) for each class. This, as a healthy side-effect improves development experience by providing type-completion and docs, though the true purpose is to define exactly what fields are exposed and useable per operation type. You can provide a ```type``` setting for any or each of your classes to limit which fields are exposed:
```
TomWFox:
```suggestion
By default, we enrich the schema by generating a number of [Input Types](https://graphql.org/learn/schema/#input-types) for each class. This, as a healthy side-effect, improves development experience by providing type-completion and docs, though the true purpose is to define exactly what fields are exposed and useable per operation type. You can provide a ```type``` setting for any or each of your classes to limit which fields are exposed:
```

TomWFox (changes requested):
Looks good, thanks for taking the time to do this! Just a few changes I can see immediately.

TomWFox (changes requested):
This new section needs to be added to the [graphql.md](https://github.com/parse-community/docs/blob/gh-pages/graphql.md) file so that it is added to the guide when the docs are built.

_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+    };
+  }>
+}
+```
+
+### Set or Update Configuration
+
+We have provided a public API in ```ParseGraphQLServer``` which accepts the above JSON object for setting (and updating) your Parse GraphQL Configuration, ```setGraphQLConfig```:
+
+```typescript
```
davimacedo: can you change to javascript?

_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+}
+```
+
+### Input Types
+
+By default, we enrich the schema by generating a number of [Input Types](https://graphql.org/learn/schema/#input-types) for each class. This, as a healthy side-effect improves development experience by providing type-completion and docs, though the true purpose is to define exactly what fields are exposed and useable per operation type. You can provide a ```type``` setting for any or each of your classes to limit which fields are exposed:
+
+In the following example, we have a custom class called ```Review``` where the fields ```rating``` and ```body``` are allowed on the ```create``` mutation, and the field ```numberOfLikes``` on the ```update``` mutation:
+
+```json
```
davimacedo: It is not a valid json and it will probably be `red` when rendered. Can you please use javascript for this one and actually all others?

_includes/graphql/customisation.md
```diff
@@ -0,0 +1,244 @@
...
+
+
+In the following example, we set the fields ```name``` and ```age``` as the only two that can be used to filter the ```_User``` class, and defining the ```createdAt``` and ```age``` fields the only sortable field whilst disabling the ascending direction on the ```createdAt``` field:
+
+```json
+{
+  classConfigs: [
+    {
+      className: "_User",
+      constraintFields: ["name", "age"],
```
davimacedo: it should be inside `type: {}` right?

davimacedo (changes requested):
Thanks for the PR! I like the way you explained. Additionally to @TomWFox comments, I added some other few suggestions. The rest seems good to me.

omairvaiyani (commented):
Added all the suggestions from above - @davimacedo I've switched everything to javascript except [line 16](https://github.com/parse-community/docs/pull/652/files#diff-2f28f36517e0808530394fe235fd3a00R13) given that it's using the `interface` structure.

davimacedo (approved):
It looks good to me

TomWFox (approved):

PR #658, opened 2019-07-31
Fix doc for creating a subscription in javascript

davimacedo:
According to [integration tests](https://github.com/parse-community/Parse-SDK-JS/blob/master/integration/test/ParseLiveQueryTest.js)


dplewis:
You should make a note that it started in 2.3.0 that it returns a Promise

davimacedo:
done

dplewis (approved):

PR #659, opened 2019-08-04
Add beta notice to the GraphQL Guide

davimacedo:
https://github.com/parse-community/parse-server/issues/5863


TomWFox (commented):
Maybe it would be good to add a note at the start of the guide, something like this...
“The GraphQL implementation is fully functional but discussions are taking place on how to improve ita implementation so there are likely to be breaking changes as new versions of Parse Server are released.”

davimacedo:
Done. Let me know what you think.

TomWFox (approved):
I like it, it provides a bit more context, we don't want people thinking it is unusable! - especially after all the hard work you've put into it!!

PR #663, opened 2019-08-08
Include more information about create and update response body

davimacedo:


_includes/rest/objects.md
```diff
@@ -99,6 +99,8 @@ The response body is a JSON object containing the `objectId` and the `createdAt`
 }
 ```
 
+The response body also includes any field that was modified using `beforeSave` and `afterSave` triggers.
```
TomWFox, line 102: Not sure about `afterSave`, the response is returned before the after save runs, right?

_includes/rest/objects.md
```diff
@@ -223,14 +225,16 @@ print result
...
-The response body is a JSON object containing just an `updatedAt` field with the timestamp of the update.
+The response body is a JSON object containing by default an `updatedAt` field with the timestamp of the update.
 
 ```json
 {
   "updatedAt": "2011-08-21T18:02:52.248Z"
 }
 ```
 
+The response body also includes any field that was modified using `beforeSave` and `afterSave` triggers.
```
TomWFox, line 236: Only the case if the `afterSave` trigger has completed - if there is a long running operation new changes may not be included.

_includes/rest/objects.md
```diff
@@ -223,14 +225,16 @@ print result
 </code></pre>
 </div>
 
-The response body is a JSON object containing just an `updatedAt` field with the timestamp of the update.
+The response body is a JSON object containing by default an `updatedAt` field with the timestamp of the update.
```
TomWFox:
```suggestion
The response body is a JSON object containing, by default, an `updatedAt` field with the timestamp of the update.
```

TomWFox (changes requested):

_includes/rest/objects.md
davimacedo: No. The response is returned after `afterSave` trigger. And in `afterSave` trigger you can change the fields that will be returned. Take a look in this test here: https://github.com/parse-community/parse-server/blob/master/spec/CloudCode.spec.js#L187

Maybe it is not well written to explain this case (the one in the test). Any suggestion?

_includes/rest/objects.md
davimacedo: same thing here

_includes/rest/objects.md
TomWFox: That test is for the `beforeSave` trigger?

Looking in the cloud code guide under [Async Behaviour](http://docs.parseplatform.org/cloudcode/guide/#async-behavior) is what gives me this impression. Is it the case that if a promise is returned or async/await is used the response is returned after the `afterSave` completes but otherwise the response is returned before the `afterSave` completes?

Sorry if I'm not making sense, I'm just migrating my cloud code to 3.0.0+ and my knowledge of promises and async/await is limited.

davimacedo: If the `afterSave` trigger returns a Promise, Parse Server will wait this promise to resolve before returning to the client and you can even change the fields that will be returned (as showed in the test case that I sent you before). If the `afterSave` trigger does not return a promise (as in the example of the docs that you sent me), the response will be sent to the client and the trigger may continue running.

TomWFox: I see, thanks for bringing me up to speed on that, I might work on making the docs more clear on that point.

TomWFox (approved):

TomWFox (changes requested):
Actually just thinking, we should probably add an explanation/link to the cloud code guide to make it clear that it depends on the configuration of the triggers - otherwise other people may get confused like me!

davimacedo:
I agree. I think we could include a section in the cloud code section showing how beforeSave and afterSave triggers could change the object fields.

TomWFox:
I’m happy to take a stab at that, but I might need some assistance with the code samples.

TomWFox:
@davimacedo I might not get round to this for a few weeks now so if you want to take this feel free.

PR #664, opened 2019-08-09
Add `cascadeSave` option in ParseObject to JS docs

RaschidJFR:
I thought it was worth adding a whole sub section about saving nested objects.

Updating docs according to feature added in parse-community/parse-sdk-js#881

### DIFF
--- a/_includes/js/objects.md
+++ b/_includes/js/objects.md
@@ -192,7 +192,7 @@ In some scenarios, you may want to prevent this default chain save. For example,
 ```javascript
 var TeamMember = Parse.Object.extend("TeamMember");
 var  = new TeamMember();
-teamMember.set('owninerAccount', ownerAccount);   // Supose `ownerAccount` to have been created earlier.
+teamMember.set('owninerAccount', ownerAccount);   // Suppose `ownerAccount` has been created earlier.
 
 teamMember.save(null, { cascadeSave: false });
 // Will save `teamMember` wihout attempting to save or modify `ownerAccount`
--- a/_includes/js/objects.md
+++ b/_includes/js/objects.md
@@ -172,6 +172,34 @@ gameScore.save({
 });
 ```
 
+### Saving Nested Objects
+You may add a `Parse.Object` as the value of a property in another `Parse.Object`. By default, when you call `save()` on the parent object, all nested objects will be created and/or saved as well in a batch operation. This feature makes really easy managing relational data as you don't have to take care of creating the objects in any specific order.
+
+```javascript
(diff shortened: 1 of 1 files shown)


_includes/js/objects.md
```diff
@@ -172,6 +172,34 @@ gameScore.save({
 });
 ```
 
+### Saving Nested Objects
+You may add a `Parse.Object` as the value of a property in another `Parse.Object`. By default, when you call `save()` on the parent object, all nested objects will be created and/or saved as well in a batch operation. This feature makes really easy managing relational data as you don't have to take care of creating the objects in any specific order.
```
TomWFox:
```suggestion
You may add a `Parse.Object` as the value of a property in another `Parse.Object`. By default, when you call `save()` on the parent object, all nested objects will be created and/or saved as well in a batch operation. This feature makes it really easy to manage relational data as you don't have to take care of creating the objects in any specific order.
```

_includes/js/objects.md
```diff
@@ -172,6 +172,34 @@ gameScore.save({
...
+
+var Parent = Parse.Object.extend("Parent");
+var parent = new Child();
+
+parent.save({child: child});
+// Automatically the object Child is created on the server
+// just before saving the Parent
+```
+
+In some scenarios, you may want to prevent this default chain save. For example, when saving a team member's profile that points to an account owned by another user to which you don't have write access. In this case, setting the option `cascadeSave` to `false` may come in handy:
```
TomWFox:
```suggestion
In some scenarios, you may want to prevent this default chain save. For example, when saving a team member's profile that points to an account owned by another user to which you don't have write access. In this case, setting the option `cascadeSave` to `false` may be useful:
```

_includes/js/objects.md
```diff
@@ -172,6 +172,34 @@ gameScore.save({
...
+
+```javascript
+var TeamMember = Parse.Object.extend("TeamMember");
+var  = new TeamMember();
+teamMember.set('owninerAccount', ownerAccount);   // Supose `ownerAccount` to have been created earlier.
```
TomWFox:
```suggestion
teamMember.set('owninerAccount', ownerAccount);   // Suppose `ownerAccount` has been created earlier.
```

TomWFox (changes requested):
Looks good! Just a couple grammatical issues I can see.

TomWFox (approved):
LGTM 👍 We'll wait until the next JS SDK release to merge this.

davimacedo (approved):

PR #675, opened 2019-09-18
Update recommended node version to 8 or newer

brodo:
If you use node 4.3, you get a lot of warnings for using a too old node version while installing parse. I found 9.11 to be a good version to use.

### DIFF
--- a/_includes/parse-server/getting-started.md
+++ b/_includes/parse-server/getting-started.md
@@ -9,7 +9,7 @@ Parse Server is an open source version of the Parse backend that can be deployed
 
 **Prerequisites**
 
-* Node 4.3
+* Node 9.11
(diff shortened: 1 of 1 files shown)


dplewis:
Currently the minimum version is 8. Our test cases run against 10.14.2 and 11.4.0.

If you have a version less than 8. You should see a message to update when you install parse server.

https://github.com/parse-community/parse-server/blob/master/postinstall.js#L42

brodo:
Ok, I've updated the PR with that info.

TomWFox (approved):
Looks good, thanks!

dplewis:
@brodo Quick question. Did you receive the error message when updating that I pointed out?

brodo:
@dplewis you mean from CI?

PR #679, opened 2019-10-28
Fix typo

priyashpatil:
Add android in available live query clients list


TomWFox (approved):
Thanks for contributing!

PR #683, opened 2019-11-14
Remove duplicate content

abehnamfard:
Using beforeSave on predefined class is explained with this title: "Predefined Classes"


TomWFox (approved):
👍 thanks for contributing!

PR #687, opened 2019-11-27
Update _live-toc.scss

### LINKED ISSUE #603: 🐛 Can't scroll docs contents nav
**Describe the bug**
When navigating the docs if I am looking at a section with a lot of subsections (i.e the [Users section in the iOS docs](https://docs.parseplatform.org/ios/guide/#users)), the navigation bar on the left side extends out of view and is not scrollable. This means if I then want to navigate to a section at the end of the docs I either have to scroll all the way down or go to a section with only a few subsections so the whole of the contents are visible.

**To Reproduce** -- __Must be using a device with a small height - I use a 13" MBP__

**Expected behavior**
I should be able to scroll the contents navigation bar on the left slide independent to the body of the docs.

**Screenshots**
  
__All of the contents are shown:__
Screenshot 2019-03-17 at 00 57 38

__Some of the contents are not visible:__
Screenshot 2019-03-17 at 01 05 39

 - OS: macOS Mojave
 - Browser: Safari

qhoirulanwar:
fix overflow on menu

### DIFF
--- a/css/lib/docs/components/_live-toc.scss
+++ b/css/lib/docs/components/_live-toc.scss
@@ -55,6 +55,9 @@
         // left:0; top:0;
         // height: 100%;
         // overflow-y: auto;
+        
+        height: 450px;
+        overflow: auto;
 
         padding: 0;
         margin: 0;
--- a/css/lib/docs/components/_live-toc.scss
+++ b/css/lib/docs/components/_live-toc.scss
@@ -56,8 +56,10 @@
         // height: 100%;
         // overflow-y: auto;
         
-        height: 450px;
+        height: 80%;
+        width: 174px;
         overflow: auto;
+        position: fixed;
 
         padding: 0;
         margin: 0;


TomWFox:
Hi, thanks for opening a pr. I'm not aware of an overflow issue, could you provide a before and after screenshot so I can see the area you are referring to?

qhoirulanwar:
**Before**
without scrolling, so we can't see the title on the menu
Before-parseserver2019-11-29 125432


**After**
after-parse-server2019-11-29 125319

TomWFox:
Right got it, this addresses #603 - a much needed fix!

TomWFox (changes requested):
I'm not keen on having a fixed height as this means that on larger screens only part of the menu can be seen without scrolling etc.

TomWFox:
Ah, I tried setting `height: auto` but that then stops the overflow scroll from working. Surely there is a better solution to this...

qhoirulanwar:
might feel better
`height: 80%; width: 174px; overflow: auto; position: fixed;`

when i use height 80% while show scrolling, i can view all list in menu, its work on larger screens & small screen

TomWFox:
That's much better! The only issue I can see is that it sticks to the bottom of the window as you start to scroll...

ScreenRecording2019-12-01at15435

igorgoncalves:
Can I try a different way? Changing the overflow to 100vh and put de ul tag inside the div. I can pull that solution with you agree

TomWFox:
@igorgoncalves if you think it will work please do! You can just submit a PR and I’ll review it.

TomWFox:
@qhoirulanwar thanks for your effort on this. It has since been addressed by @igorgoncalves so I'll close this now.

PR #696, opened 2019-12-18
Update subscribe example in live-query.md

JeffGuKang:
`query.subscribe()` => `await query.subscribe()`

```javascript
let query = new Parse.Query('People');
query.equalTo('name', 'Mengyan');
let subscription = await query.subscribe();
```


dplewis (approved):
LGTM!

PR #697, opened 2019-12-19
Adding Microsoft Graph OAuth Documentation

alann-maulana:
- [x] Microsft Graph `authData`
- [x] Example and links


alann-maulana:
ruined some file pushed, closing it..re-sending PR

PR #708, opened 2020-02-13
fix margin issue with quotes

dkreider:
Hello @parse-community!

There's a small rendering issue in the docs so figured I'd submit a PR to fix it.

Before:

before

After:

after

Any questions or feedback please let me know.


TomWFox:
Thanks for the PR!

Thats much better, could you reduce the font size also as it seems overly large to me?

dkreider:
You're very welcome @TomWFox !

I agree - the font size was rather large. Here's what it looks like now.

resized font

TomWFox (approved):
Perfect 👌

PR #714, opened 2020-02-25
Kue is no longer maintained

JeromeDeLeon:
Update the recommendation to what the maintainer of Kue recommends


TomWFox (approved):
Thanks!

PR #724, opened 2020-03-30
Fix for scroll nav guide

### LINKED ISSUE #603: 🐛 Can't scroll docs contents nav

igorgoncalves:
I just used this calc, the values are the top distances, in that way any change to $live-toc-distance-top will calculate a new heigth to ui_live_toc class

`calc(100vh - (#{$live-toc-distance-top} + 35px) - 20px);`

Closes #603

### DIFF
--- a/css/lib/docs/components/_live-toc.scss
+++ b/css/lib/docs/components/_live-toc.scss
@@ -53,7 +53,7 @@
 
         // position: fixed;
         // left:0; top:0;
-        height: 75vh;
+        height: calc(100vh - (#{$live-toc-distance-top} + 35px) - 20px);
         overflow-y: scroll;
 
         padding: 0;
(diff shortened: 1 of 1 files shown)


TomWFox (approved):
This works perfectly for me!

PR #730, opened 2020-04-18
Add context to object save in JS SDK

mtrezza:
Added description for new feature to expose Cloud Code trigger context in `Parse.Object.save()`.

---
Don't merge until

- [x] Change in JS SDK https://github.com/parse-community/Parse-SDK-JS/pull/1150
- [x] Change in Parse Server https://github.com/parse-community/parse-server/pull/6626
- [x] Replace required parse server version https://github.com/parse-community/docs/pull/730/commits/453d2fd5128c183d183b7ceaf4170b8b0833f5fb


_includes/js/objects.md
```diff
@@ -199,6 +199,29 @@ teamMember.save(null, { cascadeSave: false });
 
 ```
 
+### Cloud Code context
+
+You may pass a `context` that is accessible in Cloud Code `beforeSave` and `afterSave` triggers for that `Parse.Object`. This is useful if you want to condition certain operations in Cloud Code triggers on ephemeral information that should not be saved with the `Parse.Object` in the database. The context is ephemeral in the sense that it vanishes after the Cloud Code triggers for that particular `Parse.Object` have executed. For example:
```
TomWFox:
```suggestion
You may pass a `context` dictionary that is accessible in Cloud Code `beforeSave` and `afterSave` triggers for that `Parse.Object`. This is useful if you want to condition certain operations in Cloud Code triggers on ephemeral information that should not be saved with the `Parse.Object` in the database. The context is ephemeral in the sense that it vanishes after the Cloud Code triggers for that particular `Parse.Object` have executed. For example:
```
Or ‘`context` object’ if you prefer?

TomWFox (changes requested):
Looks good, one wee suggestion.

Also, I think it would be worth including this in the cloud code guide too. You could link to the content in the JS guide or vice versa but considering it’s a relatively small section I would just copy and paste it in.

_includes/js/objects.md
mtrezza: The context is currently in `afterSave Triggers` > `Using Request Context`.

Should I move it into its own main section, since it belongs to both triggers and beyond?

TomWFox: That sounds like a plan 👌

mtrezza: Check it out, looks much cleaner now.

- added headings per trigger group
- moved context of save triggers into own section

image

TomWFox: That’s a lot better!

TomWFox (dismissed):
Barring any changes to the implementation, this looks fab.

I love when people organise all their PRs nicely & take care of the docs early in the process, it makes it sooo much easier for us!

mtrezza:
The first thing I think of before starting a PR is "Which changes in the docs would make Tom happy?" 😌

TomWFox:
❤️ 

I think I might actually make a bot that comments that on every PR to a Parse repo 😅😝

mtrezza:
@TomWFox Do we have a guideline for adding the required minimum parse server version for a new feature to the docs?

I rarely see something like “Requires Parse Server v1.2”. Should that be added or do we expect users to find that out in the change log?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

TomWFox:
Good point, we don’t have a guideline for this but probably should.

When there is a clear section for a new feature in the docs I think it makes sense to add a note, like in some sections of the Cloud Code guide...
IMG_0901

I’ve been bitten by this when I was trying to use request context only to realise that it wasn’t included in the version I was running.

mtrezza:
Yes, I also found some of these:

image

TomWFox (approved):

mtrezza:
Uh, I lost track of that one, thanks @TomWFox

TomWFox:
No probs, me too!

PR #735, opened 2020-05-02
wrong return type.

hatimmakki:
game[@"fireProof"]` will return an `NSNumber`
should be
game[@"fireProof"]` will return an `Bool`

### DIFF
--- a/_includes/ios/objects.md
+++ b/_includes/ios/objects.md
@@ -673,7 +673,7 @@ You can access the displayName property using `armor.displayName` or `[armor dis
 ```
 </div>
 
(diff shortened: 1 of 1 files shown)


TomWFox:
I don’t know how the underlying code works here but looking at the existing explanation I wonder whether it is already correct.

It seems to suggest that the underlying value is an `NSNumber` but this is given as a bool through the `boolValue` I.e 1 = true, 0 = false.

cc @drdaz @noobs2ninjas

drdaz:
We should test this just to make sure, but given that we can access Parse objects using subscript in this way suggests to me we're probably using an NSDictionary under the hood. And they can only contain object types (not primitive types like BOOL).

The text following the excerpt in question explains that the dot getter and setter does the transformation for you, so that you get the expected type out when you access them via dot notation. 

My *guess* is that this is already correct.

noobs2ninjas:
So late to chime in on this. I'd just like to point out that subclassing avoids this extra work and might be something worth mentioning somewhere around this area.

According to [the subclassing section](https://docs.parseplatform.org/ios/guide/#properties--methods) when you subclass PFObject and use numeric primitive data types PFObject will automatically use the dynamic setters and getters to convert it to and from the internal NSNumber value. Specifically the documentation says:

> The dynamic getter will automatically extract the BOOL or int value and the dynamic setter will automatically wrap the value in an NSNumber.

PR #737, opened 2020-05-19
Update getting-started.md

AJAZSIDHIQ:
this is the right format


rogerhu (approved):

Jawnnypoo (commented):
This change isn't needed, you are not creating an `Activity` called `App`. You are creating an `Application` class called `App`. I think you are getting confused.

PR #741, opened 2020-06-14
Using the old snippet with NSError errors out.

jaysonng:
I've updated the swift section of the doc to make it work with the current version of Swift 

**Link to section:**
https://docs.parseplatform.org/ios/guide/#users

**What is the issue?**
The snippet is not valid with Swift 5.

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -58,6 +58,7 @@ To call the same Cloud function from an iOS app:
 
 ```swift
 // Swift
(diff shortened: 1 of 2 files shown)


jaysonng:
**Link to section:**
https://docs.parseplatform.org/ios/guide/#users

**What is the issue?**
The snippet is not valid with Swift 5.

TomWFox (approved):
I thought I'd updated all of these!

Thanks for the PR, looks good 👍

jaysonng:
I'll update as I encounter more. thanks Tom.

PR #743, opened 2020-07-11
removed obsolete 128KB limit of Parse Objects

mtrezza:
Removed the 128KB Parse Object size limit from the docs, as it does not exist in Open Source Parse Server.


TomWFox (approved):
Thanks!

Note to self - never assume the docs aren't outdated 😅

mtrezza:
> Note to self - never assume the docs aren't outdated 😅

Good one! Exactly the note I made to myself 😁

PR #747, opened 2020-07-22
Update objects.md of Android part

### LINKED ISSUE #745: 📙 fix guide typo
About guide, the part Objects -> Relational Data:

> By default, when fetching an object, related ParseObjects are not fetched. These objects’ values cannot be retrieved until they have been fetched like so:
```java
fetchedComment.getParseObject("post")
    .fetchIfNeededInBackground(new GetCallback<ParseObject>() {
        public void done(ParseObject post, ParseException e) {
          String title = post.getString("title");
          // Do something with your new title variable
        }
    });
```
`fetchedComment.getParseObject("post")` should be `fetchedComment.getParseObject("parent")`, right?

jhwsx:


TomWFox:
Thanks for making a PR, I just merged one with your commit which also includes the change to the iOS guide.

PR #760, opened 2020-08-10
Reflect the REQUIRED field for Google oAuth

srameshr:


SebC99:
ClientId is not required, just the google object.
And #6844 is adding the possibility to avoid even to add the google object.
So no need for this

PR #779, opened 2020-10-24
Enable Live Reload

dplewis:
Auto refreshes the browser when files are changes. No more manual refreshing!

Added in Jekyll 3.7.0


TomWFox (approved):
Nice 👌

PR #782, opened 2020-10-26
Cloud Validator Documentation

### LINKED ISSUE #781: 📙 beforeDelete Trigger Example
__Link to section:__ 
https://docs.parseplatform.org/cloudcode/guide/#delete-triggers

__What is the issue?__
The example is:

```
Parse.Cloud.beforeDelete("Album", (request) => {
  const query = new Parse.Query("Photo");
  query.equalTo("album", request.object);
  query.count()
    .then((count) => {
      if (count > 0) {
        throw "Can't delete album if it still has photos.";
    })
    .catch((error) {
      throw "Error " + error.code + " : " + error.message + " when getting photo count.";
    });
});

```
And states:

**If the function throws, the Album object will not be deleted, and the client will get an error. Otherwise, the object will be deleted normally.**

-The function isn't async, or doesn't return a promise. 
-Syntax error after the first throw (no closing bracked)
-Syntax error after .catch (no =>)
-Catching the first error will return 'Error undefined : undefined when getting photo count.'

I might be wrong, but wouldn't the delete happen prior to the error throwing?

```
Parse.Cloud.beforeDelete("Album", (request) => {
  const query = new Parse.Query("Photo");
  query.equalTo("album", request.object);
  return query.count()
    .then((count) => {
      if (count > 0) {
        throw "Can't delete album if it still has photos.";
      }
    })
});
```

or 

```
Parse.Cloud.beforeDelete("Album", async (request) => {
  const query = new Parse.Query("Photo");
  query.equalTo("album", request.object);
  const count = await query.count()
  if (count > 0) {
    throw "Can't delete album if it still has photos.";
  }
});
```

dblythy:
Documentation relating to the recently merged Cloud Validator.

For reference, here is the [PR](https://github.com/parse-community/parse-server/pull/6968).

Available in next release.

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -223,7 +223,7 @@ The full range of Built-In Validation Options are:
 
 The full range of Built-In Validation Options on `.fields` are:
 
-- `type`, the type of the `request.params.field` or `request.object.get(field)`
+- `type`, the type of the `request.params[field]` or `request.object.get(field)`
 - `default`, what the field should default to if it's `null`,
 - `required`, whether the field is required.
 - `options`, a singular option, array of options, or custom function of allowed values for the field.
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -153,7 +153,7 @@ Parse.Cloud.define("averageStars", async (request) => {
 
 If the rules specified in the validator object aren't met, the Cloud Function won't run. This means that you can confidently build your function, knowing that `request.params.movie` is defined, as well as `request.user`.
 
-## More Advanced Validation
+### More Advanced Validation
 
 *Available only on parse-server cloud code starting 4.-.-*
 
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -151,13 +151,13 @@ Parse.Cloud.define("averageStars", async (request) => {
 });
 ```
 
-If the rules specified in the validator object aren't met, the cloud function won't run. This means that you can confidently build your trigger, knowing that `request.params.movie` is defined, as well as `request.user`.
+If the rules specified in the validator object aren't met, the Cloud Function won't run. This means that you can confidently build your function, knowing that `request.params.movie` is defined, as well as `request.user`.
 
 ## More Advanced Validation
 
 *Available only on parse-server cloud code starting 4.-.-*
 
-Often, not only is it important that `request.params.movie` is defined, but also that it is the correct data type. You can do this by providing an `Object` to the `fields` parameter in the Validator.
+Often, not only is it important that `request.params.movie` is defined, but also that it's the correct data type. You can do this by providing an `Object` to the `fields` parameter in the Validator.
 
 ```javascript
 Parse.Cloud.define("averageStars", async (request) => {
@@ -213,24 +213,24 @@ Parse.Cloud.beforeSave(Parse.User, () => {
 ```
 This means that the field `accType` on `Parse.User` will be 'viewer' on signup, and will be unchangable, unless `masterKey` is provided.
 
-The full range of Built-In Validation Options are:
+The full range of built-in Validation Options are:
 
-- `requireMaster`, whether the function requires a `masterKey` to run
-- `requireUser`, whether the function requires a `request.user` to run
-- `validateMasterKey`, whether the validator should run on `masterKey` (defaults to false)
-- `fields`, an `Array` or `Object` of fields that are required on the request.
-- `requireUserKeys`, an `Array` of fields to be validated on `request.user`
+- `requireMaster`: whether the function requires a `masterKey` to run.
+- `requireUser`: whether the function requires a `request.user` to run.
+- `validateMasterKey`: whether the validator should run on `masterKey` (defaults to false).
+- `fields`: an `Array` or `Object` of fields that are required on the request.
+- `requireUserKeys`: an `Array` of fields to be validated on `request.user`.
 
-The full range of Built-In Validation Options on `.fields` are:
+The full range of built-in Validation Options on `.fields` are:
 
-- `type`, the type of the `request.params[field]` or `request.object.get(field)`
-- `default`, what the field should default to if it's `null`,
-- `required`, whether the field is required.
-- `options`, a singular option, array of options, or custom function of allowed values for the field.
-- `constant`, whether the field is immutable.
-- `error`, a custom error message if validation fails.
+- `type`: the type of the `request.params[field]` or `request.object.get(field)`.
+- `default`: what the field should default to if it's `null`.
+- `required`: whether the field is required.
+- `options`: a singular option, array of options, or custom function of allowed values for the field.
+- `constant`: whether the field is immutable.
+- `error`: a custom error message if validation fails.
 
-You can also pass a function to the Validator. This can help you apply reoccuring logic to your cloud code.
+You can also pass a function to the Validator. This can help you apply reoccuring logic to your Cloud Code.
 
 ```javascript
 const validationRules = request => {
@@ -253,10 +253,10 @@ Parse.Cloud.define('adminFunctionTwo', request => {
 ```
 
 ### Some considerations to be aware of
-- The validation function will run prior to your cloud code functions. You can use async and promises here, but try to keep the validation as simple and fast as possible so your cloud requests resolve in the shortest time.
+- The validation function will run prior to your Cloud Code Functions. You can use async and promises here, but try to keep the validation as simple and fast as possible so your cloud requests resolve quickly.
 - As previously mentioned, cloud validator objects will not validate if a masterKey is provided, unless `validateMasterKey:true` is set. However, if you set your validator to a function, the function will **always** run.
 
-This range of options should help you write more secure cloud code. If you need help in any way, feel free to reach out at our [developer supported community forum](https://community.parseplatform.org/).
+This range of options should help you write more secure Cloud Code. If you need help in any way, feel free to reach out on our [developer supported community forum](https://community.parseplatform.org/).
 
 # Cloud Jobs
 
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -24,8 +24,6 @@ Parse.Cloud.define("averageStars", async (request) => {
     sum += results[i].get("stars");
   }
   return sum / results.length;
-},{
-  fields:['movie']
 });
 ```
 
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -155,7 +155,6 @@ If the rules specified in the validator object aren't met, the Cloud Function wo
 
 ### More Advanced Validation
 
-*Available only on parse-server cloud code starting 4.-.-*
 
 Often, not only is it important that `request.params.movie` is defined, but also that it's the correct data type. You can do this by providing an `Object` to the `fields` parameter in the Validator.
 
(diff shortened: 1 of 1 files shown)


davimacedo (approved):
Nice job. Let's wait next Parse Server release and we can merge this one. Any comment here @TomWFox ?

davimacedo:
Close https://github.com/parse-community/docs/issues/781

_includes/cloudcode/cloud-code.md
```diff
@@ -24,6 +24,8 @@ Parse.Cloud.define("averageStars", async (request) => {
     sum += results[i].get("stars");
   }
   return sum / results.length;
+},{
+  fields:['movie']
```
TomWFox: I'm guessing this was a mistake...
```suggestion
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
 }
 ```
 
+## Implementing cloud function validation
+
+*Available only on parse-server cloud code starting 4.-.-*
+
+It's important to make sure the parameters provided to a Cloud function are provided, and are in the necessary format. Starting with Parse Server 4.-.-, you can now specify a validator function or object which will be called prior to your cloud function.
```
TomWFox: This doesn't read too well for me... maybe it's better like this...
```suggestion
It's important to make sure the parameters required for a Cloud function are provided, and are in the necessary format. Starting with Parse Server 4.-.-, you can now specify a validator function or object which will be called prior to your cloud function.
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+    sum += results[i].get("stars");
+  }
+  return sum / results.length;
+},{
+  fields : ['movie'],
+  requireUser: true
+});
+```
+
+If the rules specified in the validator object aren't met, the cloud function won't run. This means that you can confidently build your trigger, knowing that `request.params.movie` is defined, as well as `request.user`.
```
TomWFox:
```suggestion
If the rules specified in the validator object aren't met, the Cloud Function won't run. This means that you can confidently build your function, knowing that `request.params.movie` is defined, as well as `request.user`.
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+
+## More Advanced Validation
```
TomWFox: Should be under the 'Implementing cloud function validation' section I think...
```suggestion
### More Advanced Validation
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+
+*Available only on parse-server cloud code starting 4.-.-*
```
TomWFox: As it's mentioned above, if it's in the same section this should be unnecessary...
```suggestion
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+
+Often, not only is it important that `request.params.movie` is defined, but also that it is the correct data type. You can do this by providing an `Object` to the `fields` parameter in the Validator.
```
TomWFox:
```suggestion
Often, not only is it important that `request.params.movie` is defined, but also that it's the correct data type. You can do this by providing an `Object` to the `fields` parameter in the Validator.
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+- `requireUserKeys`, an `Array` of fields to be validated on `request.user`
+
+The full range of Built-In Validation Options on `.fields` are:
+
+- `type`, the type of the `request.params[field]` or `request.object.get(field)`
+- `default`, what the field should default to if it's `null`,
+- `required`, whether the field is required.
+- `options`, a singular option, array of options, or custom function of allowed values for the field.
+- `constant`, whether the field is immutable.
+- `error`, a custom error message if validation fails.
```
TomWFox:
```suggestion
- `type`: the type of the `request.params[field]` or `request.object.get(field)`.
- `default`: what the field should default to if it's `null`.
- `required`: whether the field is required.
- `options`: a singular option, array of options, or custom function of allowed values for the field.
- `constant`: whether the field is immutable.
- `error`: a custom error message if validation fails.
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+
+The full range of Built-In Validation Options are:
+
+- `requireMaster`, whether the function requires a `masterKey` to run
+- `requireUser`, whether the function requires a `request.user` to run
+- `validateMasterKey`, whether the validator should run on `masterKey` (defaults to false)
+- `fields`, an `Array` or `Object` of fields that are required on the request.
+- `requireUserKeys`, an `Array` of fields to be validated on `request.user`
+
+The full range of Built-In Validation Options on `.fields` are:
```
TomWFox:
```suggestion
The full range of built-in Validation Options are:

- `requireMaster`: whether the function requires a `masterKey` to run.
- `requireUser`: whether the function requires a `request.user` to run.
- `validateMasterKey`: whether the validator should run on `masterKey` (defaults to false).
- `fields`: an `Array` or `Object` of fields that are required on the request.
- `requireUserKeys`: an `Array` of fields to be validated on `request.user`.

The full range of built-in Validation Options on `.fields` are:
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+}
+
+Parse.Cloud.define('adminFunction', request => {
+// do admin code here, confident that request.user.id is masterUser, or masterKey is provided
+},validationRules)
+
+Parse.Cloud.define('adminFunctionTwo', request => {
+// do admin code here, confident that request.user.id is masterUser, or masterKey is provided
+},validationRules)
+
```
TomWFox, line 249: I imagine this is a mistake...
```suggestion
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+```
+
+### Some considerations to be aware of
+- The validation function will run prior to your cloud code functions. You can use async and promises here, but try to keep the validation as simple and fast as possible so your cloud requests resolve in the shortest time.
```
TomWFox:
```suggestion
- The validation function will run prior to your Cloud Code Functions. You can use async and promises here, but try to keep the validation as simple and fast as possible so your cloud requests resolve quickly.
```

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+- As previously mentioned, cloud validator objects will not validate if a masterKey is provided, unless `validateMasterKey:true` is set. However, if you set your validator to a function, the function will **always** run.
+
+This range of options should help you write more secure cloud code. If you need help in any way, feel free to reach out at our [developer supported community forum](https://community.parseplatform.org/).
```
TomWFox:
```suggestion
This range of options should help you write more secure Cloud Code. If you need help in any way, feel free to reach out on our [developer supported community forum](https://community.parseplatform.org/).
```

_includes/cloudcode/cloud-code.md
```diff
@@ -171,20 +304,24 @@ Viewing jobs is supported on parse-dashboard starting version 1.0.19, but you ca
...
+// do any additional beforeSave logic here
+},{
+  fields: {
+    stars : {
+      required:true,
+      options: stars => {
+        return stars >= 1 && stars =< 5;
+      },
+      error: 'Your review must be between one and five stars'
+    }
```
TomWFox, line 321: I'm not sure whether it's a good idea to completely remove examples of the 'old way' of validation before the feature is battle tested.

Perhaps here there should be another example with how the code would look with the new validation feature or a link to the section you have added above.

@davimacedo @mtrezza Any thoughts on this?

_includes/cloudcode/cloud-code.md
```diff
@@ -216,7 +353,9 @@ If you want to use `beforeSave` for a predefined class in the Parse JavaScript S
 ```javascript
 Parse.Cloud.beforeSave(Parse.User, async (request) => {
     // code here
-})
+},
+  // Validation Object or Validation Function
+)
```
TomWFox, line 355: I think it could be confusing having this here.

_includes/cloudcode/cloud-code.md
```diff
@@ -125,6 +127,137 @@ If there is an error, the response in the client looks like:
...
+The full range of Built-In Validation Options on `.fields` are:
+
+- `type`, the type of the `request.params[field]` or `request.object.get(field)`
+- `default`, what the field should default to if it's `null`,
+- `required`, whether the field is required.
+- `options`, a singular option, array of options, or custom function of allowed values for the field.
+- `constant`, whether the field is immutable.
+- `error`, a custom error message if validation fails.
+
+You can also pass a function to the Validator. This can help you apply reoccuring logic to your cloud code.
```
TomWFox:
```suggestion
You can also pass a function to the Validator. This can help you apply reoccuring logic to your Cloud Code.
```

TomWFox (changes requested):
Some small changes & thoughts but otherwise it looks great!

_includes/cloudcode/cloud-code.md
davimacedo: maybe we could should the two ways to do that

_includes/cloudcode/cloud-code.md
dblythy: I figured it might be an easy way to warm into the cloud validation. Considering `request.params.movie` is used in the query, I thought it would make sense to make sure it's not `null`.

_includes/cloudcode/cloud-code.md
dblythy: I wanted to reiterate that you can reuse `validationRules` and pass to multiple functions, but can remove!

_includes/cloudcode/cloud-code.md
dblythy: How about we leave the current method, and add the new logic below?

E.g:

```javascript
Parse.Cloud.beforeSave("Review", (request) => {
  if (request.object.get("stars") < 1) {
    throw "you cannot give less than one star";
  }

  if (request.object.get("stars") > 5) {
    throw "you cannot give more than five stars";
  }
});
```

Or, using a Cloud Validator:

```javascript
Parse.Cloud.beforeSave("Review", (request) => {
// do any additional beforeSave logic here
},{
  fields: {
    stars : {
      required:true,
      options: stars => {
        return stars >= 1 && stars =< 5;
      },
      error: 'Your review must be between one and five stars'
    }
  }
});

```

davimacedo: Yes. It works for me.

_includes/cloudcode/cloud-code.md
TomWFox: I think it's confusing because it isn't explained so if someone doesn't look at the cloud validation section they may not have any idea what it does.

_includes/cloudcode/cloud-code.md
TomWFox: Tbh I thought you had just pasted it twice by mistake (I didn't notice the functions had different names). I see where you are coming from, maybe it can stay as is.

_includes/cloudcode/cloud-code.md
TomWFox: I'm happy with that.

TomWFox (approved):
LGTM 👏

PR #789, opened 2020-11-16
Documentation for deleting files

dblythy:
Closes #726

PR #791, opened 2020-11-17
Documentation for deleting files (improvement)

mtrezza:
Improving on https://github.com/parse-community/docs/pull/789:

- Clarified wording; it could be understood that for Parse Server <4.2.0 it is not possible to delete a file using Cloud Code, which is incorrect.
- Simplified example code; it is not necessary to use a `beforeDelete` hook the example; given the `beforeDelete` name, one could assume that it is somehow related to the basic functionality of deleting a file, which it is not.


_includes/js/files.md
```diff
@@ -113,20 +113,19 @@ Parse.Cloud.httpRequest({ url: profilePhoto.url() }).then(function(response) {
...
-Parse.Cloud.beforeDelete('Profile', async (req) => {
-  const profile = req.object;
-  const profilePhoto = profile.get("photoFile");
-  await profilePhoto.destroy({ useMasterKey: true })
-});
+const profilePhoto = profile.get("photoFile");
+await profilePhoto.destroy({ useMasterKey: true });
 ```
 
+If your file is not referenced by any object in your app, it is not possible to delete it through Parse.
```
davimacedo: I believe it is actually possible to delete files not referenced by objects if you know the file name (same requirement for deleting using the REST API):

```js
(new Parse.File('fileName')).destroy();
```

mtrezza: Oh, that is interesting. I wonder if that should be a documented feature or is it more of a hack (calling destroy on an unsaved file object)? If it's supposed to be a feature, then we'd need to have a test case for this. I'll look into this when I'm on desktop.

mtrezza: I looked into Cloud Code test cases and it seems that this pattern of creating a file (without saving it) and then deleting it is used many times to test the triggers.

The REST API docs say "send a DELETE request to the files URL, postfixed with the name of the file." which is similar to creating a `ParseFile` object, setting a name (without saving it) and then calling destroy on it. While it has a "hacky" appearance, the `Parse.File.destroy()` method is doing the same as the REST API call.

However, there are some potential issues:

- The `ParseFile.destroy()` method shows that the terms "saving" and "naming" a file are used interchangeably, which could lead to confusion on the developer side or facilitate bugs in the further development of Parse Server:

```
if (!this._name) {
  throw new Error('Cannot delete an unsaved ParseFile.');
}
```

- There is no internal ParseFile state parameter that indicates whether the file has been saved or only named, but there is a change of the file name after saving a file, as the [JS SDK API reference](http://parseplatform.org/Parse-SDK-JS/api/2.18.0/Parse.File.html#name) says:
> Before save is called, this is the filename given by the user. After save is called, that name gets prefixed with a unique identifier.

That concludes that "naming" a file is not the same as "saving" a file, because "saving" actually changes the file name. Therefore I would suggest to not explicitly describe it as an official feature that a `ParseFile` object can be named and then destroyed, without ever being saved. The officially supported and "clean" process would be to create a file, keep a reference to it, and delete it by that given reference.

What do you think?

davimacedo: I believe we should change the error message `Cannot delete an unsaved ParseFile` and I'm ok in not mentioning it can be done but I'd also not state that it can not be done.

mtrezza: Sounds good, let me remove that line.

davimacedo (approved):

mtrezza:
[redacted]

_includes/js/files.md
```diff
@@ -113,20 +113,17 @@ Parse.Cloud.httpRequest({ url: profilePhoto.url() }).then(function(response) {
...
-  await profilePhoto.destroy({ useMasterKey: true })
-});
+const profilePhoto = profile.get("photoFile");
+await profilePhoto.destroy({ useMasterKey: true });
 ```
 
+* Parse Server <4.2.0
+
+The `destroy` method is available since Parse Server 4.2.0, for lower versions use the [REST API]({{ site.baseUrl }}/rest/guide/#deleting-files) to delete a file.
+
```
TomWFox: I think it's unnecessary to mention the 4.2.0 requirement twice here. I suggest removing the bullet point. Perhaps it would be nice have a note below the `Deleting Files` title. Something like this... `Available only on Parse Server Cloud Code starting 4.2.0`

TomWFox (changes requested):

_includes/js/files.md
mtrezza: > I think it's unnecessary to mention the 4.2.0 requirement twice here.

Yea, that seems redundant. How about:
* Parse Server <4.2.0
~~The `destroy` method is available since Parse Server 4.2.0, for lower versions~~ Use the REST API to delete a file.

> I suggest removing the bullet point. 

My thinking was that the docs should always reflect the newest version. Any notes regarding previous versions are becoming negligible and should therefore be below, marked with their own headline or bullet point, ordered by relevance descending. The layout would be:

< DOC FOR NEWEST VERSION, high relevance >

* Parse Server <4.2.0
< DOC FOR VERSION <4.2, medium relevance >

* Parse Server <2.2.0
< DOC FOR VERSION <2.2, low relevance >

What do you think?

TomWFox: > * Parse Server <4.2.0
>   ~The `destroy` method is available since Parse Server 4.2.0, for lower versions~ Use the REST API to delete a file.

Wording wise that works

> 
> My thinking was that the docs should always reflect the newest version. Any notes regarding previous versions are becoming negligible and should therefore be below, marked with their own headline or bullet point, ordered by relevance descending. The layout would be:
> 
> < DOC FOR NEWEST VERSION, high relevance >
> 
> * Parse Server <4.2.0
>   < DOC FOR VERSION <4.2, medium relevance >
> * Parse Server <2.2.0
>   < DOC FOR VERSION <2.2, low relevance >
> 
> What do you think?

Ahh, I didn't realise you were using the bullet point as a form of heading. I agree with the hierarchy.

I don't think the layout for the older versions works as the line(s) below aren't indented. I would suggest either having the whole doc on the bullet point or using an actual heading like `###`.

Could you also add `Available only on Parse Server Cloud Code starting 4.2.0` under neither the main `Deleting Files` heading?

mtrezza: > I don't think the layout for the older versions works as the line(s) below aren't indented. I would suggest either having the whole doc on the bullet point or using an actual heading like ###.

Right, that makes sense, I'll change that.

> Could you also add Available only on Parse Server Cloud Code starting 4.2.0 under neither the main Deleting Files heading?

When I look into the docs, I often stumble upon a "Available since parse server 2.x", and it never fails to irritate me. I think visual real estate should ideally contain only info that is useful, but these notes are not useful for maybe 99% of readers, because they use a version >2.x. Basically, these notes become pretty old pretty fast.

So instead of saying:

> ### Amazing feature
> ##### Available since Parse Server 4.2.
> Do this...


I would say:

> ### Amazing feature
> Do this...
> ##### Parse Server <4.2
> This feature is not available.

because it would be consistent with this schema:

> ### Amazing feature
> Do this...
> ##### Parse Server <4.2
> Do that instead...
> ##### Parse Server <4.0
> Do something else instead...

This way the notes are used has headlines and denote their own doc block without interfering with the section that documents the status quo.

TomWFox: I find this contradictory, you are saying that a 1 line note at the top irritates you and is not useful but also suggesting documenting potentially multiple older versions below?

Notes about features available on 2.x should probably be removed but personally I like the notes about recent releases at the top as it hopefully helps to prevent people wasting time trying to use features which don't exist on their version and helps to showcase the improvements being made across the platform.

mtrezza: I think we think the same :)

What I mean is that a line at the top "available since..." is unnecessary for most readers and therefore should be avoided because it is placed before the information the reader is actually looking for. 

A line at the bottom "on lower version use workaround..." is not as disturbing since it is placed below the information most readers are looking for.

Now, whether it should be mentioned that a feature is not available or that there is a workaround for lower versions is a different question. I think it should, and there may be better forms of denoting the availability of features.

TomWFox: Sorry for the lack of reply!

I'm not sure I agree with your assumptions about 'most readers' but I really don't have the head space to discuss this further at the moment.

I'll merge the PR as it's definitely as improvement regardless!

TomWFox (approved):

PR #797, opened 2020-11-25
Fix casing for afterLiveQueryEvent

dblythy:
Pending next release, `request.event` on `afterLiveQueryEvent` will be lowercase for consistency with the SDKs.

It's also worth noting that the example code in the docs won't work for the current parse server version as `Update` has a capitalised "U"

`if (request.event != "update") {`

Should we mention that this change is upcoming?

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -815,7 +815,7 @@ Parse.Cloud.afterLiveQueryEvent('MyObject', async (request) => {
 
 // Extend matchesQuery functionality to LiveQuery
 Parse.Cloud.afterLiveQueryEvent('MyObject', async (request) => {
-  if (request.event != "Create") {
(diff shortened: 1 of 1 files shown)


TomWFox (approved):

TomWFox:
It seems a bit OTT to mention the version required etc for such a small change but 🤷‍♂️

dblythy:
Yeah, agree. I'm just concerned about people that are already using this feature, that it won't work the way they expect it when they update.

mtrezza:
If this is considered a bug (which I think it should be) that has now been corrected, it would be enough to mention this as a breaking change in the release notes. Bug fixes are usually not mentioned in the docs.

TomWFox:
I think @mtrezza is right. It's difficult but I think we have to draw the line somewhere.

dblythy:
That is cool, thanks for the discussion.

PR #800, opened 2020-11-27
Delete compatibility.md

alyssoncm:
Removing not relevant/outdated Parse.com session


mtrezza:
Thanks for this PR, I think this document is indeed obsolete.
Did you check whether this document is liked from somewhere else in the docs, maybe from a menu?

alyssoncm:
I haven't. Let me do it

alyssoncm:
Just found a link on the menu. Also looked for external references on google and did not found

TomWFox:
I'll try and have a look at this over the weekend but my initial thought is that when thinking about removing this doc before I felt that some sections still provide value even though they are irrelevant in the context of compatibility with Parse.com.

I'll pull up any specific examples when I review but just thought I'd mention it now incase anyone else has thoughts on that.

mtrezza:
It could be helpful to mention the depreciation of `Parse.User.current()` and `Parse.Cloud.useMasterKey()` because when coming across code examples from hosted Parse.com, it makes it easier to understand why they don't work on Parse Server.

Not sure if we need a separate document for that though, maybe mentioning this in the code section for Parse User or Cloud Code respectively would be sufficient.

TomWFox:
I've looked though now, my only concern was removing the section email adapters but there is an exact copy of that in the Parse Server readme so I'm ok with that.

>  could be helpful to mention the depreciation of Parse.User.current() and Parse.Cloud.useMasterKey() because when coming across code examples from hosted Parse.com

If someone sees outdated code examples and then check our docs, they should find the correct code so I'm not sure this is much of an issue. In the cloud code guide there is a [section on master key use](https://docs.parseplatform.org/cloudcode/guide/#using-the-master-key-in-cloud-code). There is mention of `request.user` some way down [this section](https://docs.parseplatform.org/cloudcode/guide/#cloud-functions) - perhaps could be more prominent but it is used in lots of the code examples.

TomWFox (approved):

mtrezza:
Sounds good to me.

TomWFox:
Nice to finally get that removed, thanks Alysson!

PR #820, opened 2021-03-06
Update keys and excludeKeys to use arrays

cbaker6:
Supports changes made in pending [PR](https://github.com/parse-community/parse-server/pull/7242). Can see [here](https://github.com/parse-community/parse-server/issues/7245) for further discussion.

### DIFF
--- a/_includes/rest/queries.md
+++ b/_includes/rest/queries.md
@@ -414,7 +414,7 @@ print result
 
 You can restrict the fields returned by passing `keys` or `excludeKeys` as an [array](#arrays). To retrieve documents that contain only the `score` and `playerName` fields (and also special built-in fields such as `objectId`, `createdAt`, and `updatedAt`):
 
-* On Parse Server <=4.5.0 pass a comma-delimited string, e.g. `"score,playerName"` instead of an array for `keys` and `excludeKeys`.
+* On Parse Server <5.0.0 pass a comma-delimited string, e.g. `"score,playerName"` instead of an array for `keys` and `excludeKeys`.
 
 <div class="language-toggle">
 <pre><code class="bash">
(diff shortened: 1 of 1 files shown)


cbaker6:
@TomWFox Assuming the linked PR gets accepted, not sure how you want to handle/note that if someone uses the changes (array of keys) instead of a list of keys on a Parse Server < 5.0.0, the query will return all keys. If they use either a list or an array on a Parse Server >= 5.0.0, they will receive what they are expecting.

_includes/rest/queries.md
```diff
@@ -412,7 +412,7 @@ print result
 </code></pre>
 </div>
 
-You can restrict the fields returned by passing `keys` or `excludeKeys` a comma-separated list. To retrieve documents that contain only the `score` and `playerName` fields (and also special built-in fields such as `objectId`, `createdAt`, and `updatedAt`):
+You can restrict the fields returned by passing `keys` or `excludeKeys` as an [array](#arrays). To retrieve documents that contain only the `score` and `playerName` fields (and also special built-in fields such as `objectId`, `createdAt`, and `updatedAt`):
```
mtrezza, line 415: I think it's good not mention that *both* will be possible after the PR, because we want to [phase out](https://github.com/parse-community/parse-server/issues/7245#issuecomment-791978838) the string variant. For Parse Server <=4.5.0 I suggest to just add a note like:
> On Parse Server <=4.5.0 pass a comma-delimited string instead of an array.

In a format similar to what we already have:
image

cbaker6: @mtrezza How do you suggest to address this? If the PR is merged, the use of arrays won't be applicable to Parse Server <= 4.5.0, only newer, they have to use comma delineated strings.

I can see a dev saying, "it doesn't work", because they are using an older version of the server.

mtrezza: You are reading my comments faster than I can write them :-) Please see my update above.

cbaker6: This works for me. Lets see if @TomWFox thinks we need to add anything more

TomWFox: Sounds good to me 👍

cbaker6:
@mtrezza I believe this can be merged now since the related PR is released on the server

PR #821, opened 2021-04-04
Add afterFind Considerations

dblythy:
Adds considerations for using `fetch` in an `afterFind` trigger.

### DIFF
--- a/_includes/cloudcode/cloud-code.md
+++ b/_includes/cloudcode/cloud-code.md
@@ -673,7 +673,7 @@ Parse.Cloud.afterFind(Parse.User, async (request) => {
 ```
 
 ### Some considerations to be aware of
-- If you use the `masterKey` to fetch a pointer in an `afterFind` trigger, it will be sent in full to the client. Be sure to check that the returned objects and pointers are secured prior to returning to the client.
+- If you use the `masterKey` to fetch a pointer in an `afterFind` trigger, it will be sent in full to the client. Prior to returning to the client, be sure to check that the returned objects and pointers do not contain information that the client should not be able to access
 
 # Session Triggers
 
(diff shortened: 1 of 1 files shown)


davimacedo (approved):
It looks good to me. @TomWFox any comment?

_includes/cloudcode/cloud-code.md
```diff
@@ -672,6 +672,9 @@ Parse.Cloud.afterFind(Parse.User, async (request) => {
 })
 ```
 
+### Some considerations to be aware of
+- If you use the `masterKey` to fetch a pointer in an `afterFind` trigger, it will be sent in full to the client. Be sure to check that the returned objects and pointers are secured prior to returning to the client.
```
TomWFox: "the returned objects and pointers are secured" this is a bit ambiguous, I think 'secured' could make people think they should be changing ACLs etc.

Perhaps something like "Prior to returning to the client, be sure to check that the returned objects and pointers do not contain information that the client should not be able to access."

TomWFox (approved):
LGTM 👍, thanks for the PR!

_includes/cloudcode/cloud-code.md
```diff
@@ -672,6 +672,9 @@ Parse.Cloud.afterFind(Parse.User, async (request) => {
 })
 ```
 
+### Some considerations to be aware of
```
mtrezza, line 675: If we want to make developers aware of security implications, I think we should make a broader change in the docs.

1) Move this to the already existing `Using the Master Key in cloud code` section
2) Refactor the existing section to make it more prominent
3) Articulate the example in a more general way, because it does not only apply to afterFind and pointers

Changed in https://github.com/parse-community/docs/pull/823

mtrezza (commented):
A suggestion

PR #822, opened 2021-04-07
Removed MongoRocks from the parse-server guide

### LINKED ISSUE #718: 📙 remove info about mongorocks
__Link to section:__
Please provide a specific link to the section in question. In the docs, you can click on a heading and then copy the relative link from the address bar.

__What is the issue?__
Parse-community/parse-server#6468

hej2010:
Issue #718 and from [here](https://community.parseplatform.org/t/using-mongodb-with-rocksdb/1543/6)

closes #718

### DIFF
--- a/_includes/parse-server/database.md
+++ b/_includes/parse-server/database.md
@@ -17,8 +17,6 @@ If this is your first time setting up a MongoDB instance, we recommend a Databas
 
 When using MongoDB with your Parse app, you need to manage your indexes yourself. You will also need to size up your database as your data grows.
 
-If you are planning to run MongoDB on your own infrastructure, we highly recommend using the [RocksDB Storage Engine](#using-mongodb--rocksdb).
-
 In order to allow for better scaling of your data layer, it is possible to direct queries to a mongodb secondary for read operations.  See: [Mongo Read Preference](#using-mongodb-read-preference).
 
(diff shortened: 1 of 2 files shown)


mtrezza:
Thanks for opening this PR!

Did you check whether there are any links to that file that should also be removed?

hej2010:
Now I did :) No other references to it that I can find

mtrezza:
Thanks!

@TomWFox How is https://docs.parseplatform.org/parse-server/guide/#using-mongodb--rocksdb composed? Is there another reference that needs to be removed to not make it appear in the table of contents?

TomWFox:
That should do the trick as is, thanks for the PR @hej2010!

TomWFox:
Should we be adding in a note about WiredTiger as well as removing the note about MongoRocks?

mtrezza:
The current Parse Server version still supports both storage engines of MongoDB (MMAPv1, WiredTiger).

The next version of Parse Server will drop support for MongoDB 3.6 and Node 8 - 10. In fact we should remove these requirements because they change every year and we have a compatibility table in the Parse Server README, which makes more sense.

TomWFox (approved):
Looks good to me then 👍

mtrezza (approved):

PR #823, opened 2021-04-08
Improve security section

mtrezza:
Based on https://github.com/parse-community/docs/pull/821, this PR
- moves the added security note regarding master key into the existing master key section
- refactors the section as `Security` section to give it more prominence


mtrezza:
@dblythy What do you think?

_includes/cloudcode/cloud-code.md
```diff
@@ -871,13 +867,23 @@ To learn more, read the [Parse LiveQuery Protocol Specification](https://github.
...
 
 ```javascript
-query.find({useMasterKey:true});
-object.save(null,{useMasterKey:true});
-Parse.Object.saveAll(objects,{useMasterKey:true});
+Parse.Object.saveAll(objects, { useMasterKey: true });
 ```
+
+### Considerations
+- If you set `masterKey: true` when fetching objects with a query or relation in [Cloud Functions](https://docs.parseplatform.org/cloudcode/guide/#cloud-functions) or [Find Triggers](https://docs.parseplatform.org/cloudcode/guide/#find-triggers), the complete object will be returned. You may want to remove object properties that the client should not be able to access before sending it to the client.
```
dblythy: Use `{{ site.baseUrl }}/cloudcode/guide/#cloud-functions`

dblythy:
I think this an improvement. I think this is a good edition as I think this concern is probably relevant to returning objects from cloud functions, such as:

```
Parse.Cloud.define('getStats', async () => {
  const stat = await new Parse.Query('SecureStat').first({useMasterKey:true});
  return stat; // insecure returns whole object (I think)
  return stat.get('count') // only return relevant fields
});
```

 I like the idea of a security section and think it will continue to grow as we continue to evolve the docs to recommend best practice guidelines.

_includes/cloudcode/cloud-code.md
mtrezza: Done

TomWFox (approved):
LGMT 👍

PR #825, opened 2021-04-30
add s3 adapter examples for Linode and Backblaze

alioguzhan:
I upated the docs as discussed [here](https://github.com/parse-community/parse-server-s3-adapter/issues/141) and [here](https://github.com/parse-community/parse-server/pull/7300).

### DIFF
--- a/_includes/parse-server/file-adapters.md
+++ b/_includes/parse-server/file-adapters.md
@@ -175,80 +175,6 @@ var api = new ParseServer({
 
 Don't forget to change **S3_ACCESS_KEY**, **S3_SECRET_KEY** and **S3_BUCKET** to their correct value.
 
-##### S3Adapter constructor options
-
-```js
-new S3Adapter(accessKey, secretKey, bucket, options)
-```
-
-| Parameter | Description | Notes |
-| --------- | ----------- | ----- |
-| accessKey    | The AWS access key for a user that has the required permissions | Required. |
-| secretKey    | The AWS secret key for the user | Required. |
-| bucket       | The name of your S3 bucket. | Required. |
-| options      | JavaScript object (map) that can contain: | |
-| region       | Key in `options`. Set the AWS region to connect to. | Optional. Default: `us-east-1` |
-| bucketPrefix | Key in `options`. Set to create all the files with the specified prefix added to the filename. Can be used to put all the files for an app in a folder with 'folder/'. | Optional. Default: `null` |
-| directAccess | Key in `options`. Controls whether reads are going directly to S3 or proxied through your Parse Server. If set to true, files will be made publicly accessible, and reads will not be proxied. | Optional. Default: `false` |
-| baseUrl | Key in `options`. The base URL the file adapter uses to determine the file location for direct access. | Optional. Default: `null`. To be used when `directAccess=true`. When set the file adapter returns a file URL in format `baseUrl/bucketPrefix` + `filename`. Example for `baseUrl='http://domain.com/folder'` and `bucketPrefix='prefix_'` the returned file location is  `http://domain.com/folder/prefix_file.txt`. |
-| baseUrlDirect | Key in `options`. Is `true` if the file adapter should ignore the bucket prefix when determining the file location for direct access. | Optional. Default: `false`. To be used when `directAccess=true` and `baseUrl` is set. When set to `true`, the file adapter returns a file URL in format `baseUrl/filename`. Example for `baseUrl='http://domain.com/folder'` and `baseUrlDirect=true` the returned file location is `http://domain.com/folder/file.txt`. |
-| globalCacheControl | Key in `options`. The `Cache-Control` http header to set in the file request. | Optional. Default: `null`. Example: `public, max-age=86400` for 24 hrs caching. More info [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.1). |
-
-
-## Configuring `GCSAdapter`
-
-Unlike the S3 adapter, you must create a new Cloud Storage bucket, as this is not created automatically. See the Google Cloud guide on [Authentication](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/master/guides/authentication) for more details.
-
-To generate a private key in the Cloud Platform Console follow [these instructions](https://cloud.google.com/storage/docs/authentication#generating-a-private-key).
-
-### Installation
-
-Starting 2.2.6, GCS Adapter is not provided by default by parse-server. To install run:
-
-```
-npm install --save parse-server-gcs-adapter
-```
-
-### Configuration options
-
-Writing to your Google Cloud Storage bucket from Parse Server is as simple as configuring and using the GCS files adapter.
-
-#### Using environment variables
-
-You can use Google Cloud Storage to host your static files by setting the following environment variables:
-
-| Variable Name | Description | Notes |
-| ------------- | ----------- | ----- |
-| PARSE_SERVER_FILES_ADAPTER  | Set this variable to 'parse-server-gcs-adapter'. | Required. |
-| GCP_PROJECT_ID              | The project ID from the Google Developer's Console. | Required. |
-| GCP_KEYFILE_PATH            | Full path to the a .json, .pem, or .p12 key downloaded from the Google Developers Console. | Required. |
-| GCS_BUCKET                  | The name of your GCS bucket. | Required. |
-| GCS_BUCKET_PREFIX           | Create all the files with the specified prefix added to the filename. Can be used to put all the files for an app in a folder with 'folder/'. | Optional. |
-| GCS_DIRECT_ACCESS           | Whether reads are going directly to GCS or proxied through your Parse Server. | Optional. Default: false |
-
-#### Passing as options
-
-If you're using Node.js/Express:
-
-```javascript
-...
-var GCSAdapter = require('parse-server-gcs-adapter');
-
-var api = new ParseServer({
-  databaseURI: databaseUri || 'mongodb://localhost:27017/dev',
-  appId: process.env.APP_ID || 'APPLICATION_ID',
-  masterKey: process.env.MASTER_KEY || 'MASTER_KEY',
-  ...
-  filesAdapter: new GCSAdapter(
-    "GCP_PROJECT_ID",
-    "GCP_KEYFILE_PATH",
-    "GCS_BUCKET",
-    {directAccess: true}
-  ),
-  ...
-});
-```
-
 ##### S3Adapter configuration for Digital Ocean Spaces
 
 Spaces is an S3 equivalent prodivided by Digital Ocean. It's use the same api as S3 so you can use it with the S3 Adapter.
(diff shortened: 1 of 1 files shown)


alioguzhan:
Hey @mtrezza, 

You can check if this looks right or not. 

Thanks

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
 
 Don't forget to change **S3_ACCESS_KEY**, **S3_SECRET_KEY** and **S3_BUCKET** to their correct value.
 
+##### S3Adapter configuration for Digital Ocean Spaces
```
mtrezza: I think this needs another heading level or the whole chapter needs another index structure; its really hard to find that there are config examples for S3-compatible storage services in the ToC.

Also, shorten headline; it is already clear that these are S3 Adapter options:

> Options for Digital Ocean Spaces

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+
+Spaces is an S3 equivalent prodivided by Digital Ocean. It's use the same api as S3 so you can use it with the S3 Adapter.
```
mtrezza: It's not an equivalent (that would be judgmental); let's stick with "S3-compatible" for all S3-compatible services here.

> [Digital Ocean Spaces](https://try.digitalocean.com/cloud-storage) is a S3-compatible storage service in the cloud. See their [documentation](https://docs.digitalocean.com/products/spaces/) for more details.

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+You just need to change the AWS Endpoint to point to your Spaces endpoint.
+
+```javascript
+...
+var S3Adapter = require('parse-server').S3Adapter;
+var AWS = require("aws-sdk");
```
mtrezza: Why do we need to require the AWS SDK?

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+
+//Set Digital Ocean Spaces EndPoint
+const spacesEndpoint = new AWS.Endpoint(process.env.SPACES_ENDPOINT);
+//Define S3 options
+var s3Options = {
```
mtrezza: `const s3Options ...`

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+  bucket: process.env.SPACES_BUCKET_NAME,
```
mtrezza: Remove `process.env.` from all variables, this is specific solution we don't get get into in this example for the sake of compactness.

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+  baseUrl: process.env.SPACES_BASE_URL,
+  region: process.env.SPACES_REGION,
+  directAccess: true,
+  globalCacheControl: "public, max-age=31536000",
+  bucketPrefix: process.env.SPACES_BUCKET_PREFIX,
+  s3overrides: {
+    accessKeyId: process.env.SPACES_ACCESS_KEY,
+    secretAccessKey: process.env.SPACES_SECRET_KEY,
```
mtrezza: This is obviously bad practice, but without `process.env.` it is at least move ambivalent without suggesting something unsafe.

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
+var s3Adapter = new S3Adapter(s3Options);
+
+var api = new ParseServer({
+  databaseURI: databaseUri || 'mongodb://localhost:27017/dev',
+  appId: process.env.APP_ID || 'APPLICATION_ID',
+  masterKey: process.env.MASTER_KEY || 'MASTER_KEY',
+  ...
+  filesAdapter: s3Adapter
+  ...
+});
```
mtrezza: This is not needed, we only need to demonstrate the service specific adapter options.

_includes/parse-server/file-adapters.md
```diff
@@ -175,6 +175,94 @@ var api = new ParseServer({
...
 Don't forget to change **S3_ACCESS_KEY**, **S3_SECRET_KEY** and **S3_BUCKET** to their correct value.
 
+##### S3Adapter configuration for Digital Ocean Spaces
+
+Spaces is an S3 equivalent prodivided by Digital Ocean. It's use the same api as S3 so you can use it with the S3 Adapter.
+You just need to change the AWS Endpoint to point to your Spaces endpoint.
+
+```javascript
+...
+var S3Adapter = require('parse-server').S3Adapter;
```
mtrezza: This is not needed, we only need to demonstrate the service specific adapter options.

mtrezza (changes requested):
Thanks for the changes, the comments I added for DO Spaces should also be made in the same way for the other 2 S3-compatible services.

alioguzhan:
I updated the heading levels. Below is from my local instance. I followed the same pattern as `GridStoreAdapter`. Now we can see the configs for 3rd-party providers in the ToC section. What do you think? @mtrezza 
 
image

_includes/parse-server/file-adapters.md
alioguzhan: Oops. It turns out, you are right. I thought we need to process the 3rd-party provider's endpoint with `aws-sdk` to make it work. But it is working when I just pass it to adapter options as it is without the `aws-sdk`. Weird. Almost all examples on the internet have this aws-sdk thing. Even the README of [parse-s3-adapter](https://github.com/parse-community/parse-server-s3-adapter#usage-with-digital-ocean-spaces). Do you have any idea why everyone prefers to implement this through AWS SDK?

mtrezza:
Can you do change the heading levels to this:
- S3-compatible services
  - Digital Ocean Spaces
  - ...

Then maybe the "Digital Ocean Spaces" won't show up in the index, but the umbrella headline leads to it. Plus it makes the headlines more concise.

Alternatively, if you prefer to see the services in the index, you could just remove the "Options for" from each existing headline, because that is rather repetitive. I personally find that docs difficult to scroll through due to the somewhat antiquated design, so it may make sense to keep the headlines in the index.

_includes/parse-server/file-adapters.md
alioguzhan: Just asking, Why this is obviously bad practice? What is the alternative option for this? @mtrezza

alioguzhan:
This is ready for another review, btw. Thanks @mtrezza

_includes/parse-server/file-adapters.md
```diff
@@ -175,7 +175,67 @@ var api = new ParseServer({
...
+  directAccess: false,
+  s3overrides: {
+    accessKeyId: "S3_ACCESS_KEY", // bucket access key
+    secretAccessKey: "S3_SECRET_KEY", // bucket secret key
+    endpoint: "S3_ENDPOINT", // regionName.linodeobjects.com
+  },
+};
+```
+
+### Backblaze
```
mtrezza: I'm not sure that is the correct heading level? How can `Backblaze` (and the other 2) have a higher level than `S3Adapter constructor options`? From this is looks like `S3Adapter constructor options` was a sub of  `Backblaze`.

_includes/parse-server/file-adapters.md
mtrezza: Making secrets accessible via env vars is bad practice because they can be easier compromised.

_includes/parse-server/file-adapters.md
mtrezza: It's not needed, its part of the adapter.

_includes/parse-server/file-adapters.md
mtrezza: This is not resolved but I'll comment on the new changes.

_includes/parse-server/file-adapters.md
```diff
@@ -175,7 +175,67 @@ var api = new ParseServer({
 
 Don't forget to change **S3_ACCESS_KEY**, **S3_SECRET_KEY** and **S3_BUCKET** to their correct value.
 
-##### S3Adapter constructor options
+### Digital Ocean Spaces
+
+Spaces is an S3-compatible provided by Digital Ocean. It's use the same api as S3 so you can use it with the S3 Adapter.
```
mtrezza: Let's keep the descriptions uniform and containing the same info (link to product and docs). Please change to:

> [Digital Ocean Spaces](https://try.digitalocean.com/cloud-storage) is a S3-compatible storage service in the cloud. See their [documentation](https://docs.digitalocean.com/products/spaces/) for more details.

_includes/parse-server/file-adapters.md
```diff
@@ -175,7 +175,67 @@ var api = new ParseServer({
...
+    accessKeyId: "SPACES_ACCESS_KEY",
+    secretAccessKey: "SPACES_SECRET_KEY",
+    endpoint: 'SPACES_ENDPOINT'
+  }
+};
+```
+
+### Linode Object Storage
+
+Object Storage is an S3-compatible storage service from Linode. We can configure our S3Adapter to use Linode's service. Please refer to [this guide](https://www.linode.com/docs/guides/how-to-use-object-storage/) for more details on Linode's API.
```
mtrezza: Please change to (include links):

> [Linode Object Storage](...) is a S3-compatible storage service in the cloud. See their [documentation](...) for more details.

_includes/parse-server/file-adapters.md
```diff
@@ -175,7 +175,67 @@ var api = new ParseServer({
...
+    accessKeyId: "S3_ACCESS_KEY", // bucket access key
+    secretAccessKey: "S3_SECRET_KEY", // bucket secret key
+    endpoint: "S3_ENDPOINT", // regionName.linodeobjects.com
+  },
+};
+```
+
+### Backblaze
+
+We also can use Backblaze's S3-compatible [B2 Cloud Storage](https://www.backblaze.com/b2/cloud-storage.html) as a storage adapter. Here is a working configuration for B2:
```
mtrezza: Please change to (include links):

> [Backblaze B2 Cloud Storage](...) is a S3-compatible storage service in the cloud. See their [documentation](...) for more details.

_includes/parse-server/file-adapters.md
mtrezza: Please change heading to include product name: `Backblaze B2 Cloud Storage`

mtrezza (changes requested):

_includes/parse-server/file-adapters.md
```diff
@@ -175,7 +175,67 @@ var api = new ParseServer({
...
+  region: 'us-west-000',
+  s3overrides: {
+    endpoint: "S3_ENDPOINT", // check backblaze bucket endpoint
+    accessKeyId: "S3_ACCESS_KEY",
+    secretAccessKey: "S3_SECRET_KEY"
+  },
+};
+```
+
+#### S3Adapter constructor options
```
mtrezza: Can you please change this headline to simply `Adapter Options`?

_includes/parse-server/file-adapters.md
alioguzhan: You are right. Missed that. I like to keep the s3 providers' section as it is now. What if I lift `constructor options` up one more level? It will stay at the same level as others. Or do you have any other suggestions? I can implement that, too.

alioguzhan: image

mtrezza: Shouldn't the s3 compatible headlines be subs of the Adapter Option headline? Now that we reduced that to the options object, these are really only options examples.

mtrezza: Then I suggest to add the headlines like this, it's probably the easiest:
- Adapter Options
- S3-compatible Services
  - Digital Ocean Spaces
  - ...

That means adding the option examples after the adapter options and summarizing them under a common headline.

_includes/parse-server/file-adapters.md
alioguzhan: OK, but what would be an alternative way to pass those secrets to my express app code? I can hard-code those into code or read them from the external resources which is something like environment variables. I am just trying to understand the best practice here. @mtrezza

_includes/parse-server/file-adapters.md
alioguzhan: Done and pushed. Can you please check it? Thanks

mtrezza:
I think this ready for review by @TomWFox

TomWFox (approved):
LGTM, thanks for the PR @alioguzhan! 🙌

I'm assuming these changes are based on https://github.com/parse-community/parse-server/pull/7300, so we'll wait for the next Parse Server release to merge this PR?

alioguzhan:
@TomWFox  Yes, we will wait, I guess.

@mtrezza it appears there is still a pending change request which blocks merging. Can you check or resolve that? I am not authorized I believe. Thanks.

_includes/parse-server/file-adapters.md
```diff
@@ -194,6 +194,66 @@ new S3Adapter(accessKey, secretKey, bucket, options)
...
+#### Digital Ocean Spaces
+
+[Digital Ocean Spaces](https://try.digitalocean.com/cloud-storage) is an S3-compatible storage service in the cloud. See their [documentation](https://docs.digitalocean.com/products/spaces/) for more details.
+
+```javascript
+const s3Options = {
+  bucket: "SPACES_BUCKET_NAME",
+  baseUrl: "SPACES_BASE_URL",
+  region: "SPACES_REGION",
+  directAccess: true,
```
mtrezza: Are directAccess and cacheControl necessary for these examples? I think we should only keep the essential settings here, otherwise readers may think these parameters are necessary for the service.

alioguzhan: You are right. Updated.

_includes/parse-server/file-adapters.md
TomWFox: I was curious about this and found a [good piece weighing up the different options](https://medium.com/@GaryHarrower/safely-managing-secrets-with-node-js-and-google-secret-manager-1d9a1055b9e4). Not sure what the right think to do here is tho.

alioguzhan: Well, this is all new for me. Using environment variables for my secrets was the least suspicious thing in my entire setup for years. I think now it is time to digging to learn more. Thanks for the article btw. @TomWFox

TomWFox: No problem, new for me too! 🙂

mtrezza:
@alioguzhan It's been some time, is this ready to get merged?

alioguzhan:
@mtrezza Yes, We have been waiting for version `5.x`  to merge this. Since it is out, we can merge this.

PR #830, opened 2021-05-31
Need session activation from Parse

FandryNoutah:
After logging in with parseUser, it can be fine if you start session automatically and developers don't need to call PHP session_start() function, or define into the documentation the full use of the session, thanks.

### DIFF
--- a/_includes/php/users.md
+++ b/_includes/php/users.md
@@ -88,10 +88,9 @@ Note that this will only send if the account for the email requested has not alr
 
 It would be bothersome if the user had to log in every time they open your app. You can avoid this by using the cached current `ParseUser` object.
 
-By default, whenever you use any signup or login methods, the user will be saved in PHP Session storage (The `$_SESSION` superglobal.)
+By default, whenever you use any signup or login methods, the user will be saved in PHP Session storage, the `$_SESSION` superglobal. When re-initializing the Parse PHP SDK, you may need to instruct PHP to load the stored session data into the `$_SESSION` superglobal using `session_start()` before initializing the Parse PHP SDK to make Parse use persistent storage and continue the previous user session.
 
 ```php
-//Here we need to call session_start() if we want to rerieve the current user
 $currentUser = ParseUser::getCurrentUser();
 if ($currentUser) {
     // do stuff with the user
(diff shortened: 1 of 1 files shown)


mtrezza:
@TomWFox Did you request a technical review? This change has errors in style and typos.

TomWFox:
Hmm, perhaps technical and style. A lack of technical understanding on my part (never written any php!) also leaves me unsure of the best style here!

mtrezza:
@dplewis As I understand it, the developer needs to call `session_start()` to make PHP read the session into the `$_SESSION` superglobal, which then becomes available to Parse when initializing the SDK to resume the session. Do you know why we let the developer do this manually in the PHP SDK instead of doing this automatically? For comparison, how is that handled in the JS SDK?

_includes/php/users.md
```diff
@@ -91,6 +91,7 @@ It would be bothersome if the user had to log in every time they open your app.
 By default, whenever you use any signup or login methods, the user will be saved in PHP Session storage (The `$_SESSION` superglobal.)
 
 ```php
+//Here we need to call session_start() if we want to rerieve the current user
```
mtrezza: Can you please move the line out of the code block and add it after the sentence above?

> By default, whenever you use any signup or login methods, the user will be saved in PHP Session storage, the `$_SESSION` superglobal. When re-initializing the Parse PHP SDK, you may need to instruct PHP to load the stored session data into the `$_SESSION` superglobal using `session_start()` before initializing the Parse PHP SDK to make Parse continue the previous user session.

mtrezza (changes requested):

dplewis:
@mtrezza I believe that calling `session_start()` multiple times will result in an error being thrown which is probably why we don't include it in the SDK. 

If a developer tries to use persistant storage an error is called
```
'PHP session_start() must be called first.'
```
If a developer only wants to query and such without persistant storage then calling `session_start()` in the SDK will be a performance hit.

A note in the docs is fine, it should have been added a while ago, see
https://github.com/parse-community/parse-php-sdk/issues/41

mtrezza:
Thanks for that insight, and that you even found https://github.com/parse-community/parse-php-sdk/issues/41. Note is in the docs. We should be good now.

mtrezza (approved):

PR #832, opened 2021-06-02
Parse Server 6641: document the new /loginAs REST endpoint

GormanFletcher:
This PR adds REST docs for the new `/loginAs` endpoint, as discussed in [Parse Server #7406](https://github.com/parse-community/parse-server/pull/7406), which implements the request from [Parse Server #6641](https://github.com/parse-community/parse-server/issues/6641).


_includes/rest/users.md
```diff
@@ -613,3 +611,53 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+
+## Impersonating a user
+
+An application may have use cases where an operator or automated system needs to
+take action on a user's behalf, under the user's authority, without access to
+the user's credentials. The Parse REST API supports these use cases with the
+`/loginAs` endpoint. This endpoint takes a `userId` parameter, and must be
+called using the master key. It will create a session for the given user ID, and
+will return the same response format as the `/login` endpoint. The new session
+will have a `createdWith` value of `create`, and an `authProvider` value of
```
mtrezza: > `createdWith` value of `create`

Is that a typo? I think the key is `action`. You could simply say:

> The created session has the property `createdWith: { action: login, authProvider: masterkey }`.

_includes/rest/users.md
```diff
@@ -613,3 +611,53 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+`masterkey`.
+
+**The `/loginAs` endpoint performs no identity verification** - any client with
+the master key can log in as any user. You are strongly encouraged to apply
+careful security around any systems that call the `/loginAs` endpoint.
```
mtrezza: It sounds a bit misleading to me. There is the requirement of master key, so the endpoint is secured, like any other op that requires the master key. The paragraph sounds a bit as if the developer had to secure the endpoint manually, e.g. with a firewall from external access.

_includes/rest/users.md
```diff
@@ -613,3 +611,53 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+       "X-Parse-Master-Key": "${MASTER_KEY}",
+       "X-Parse-Revocable-Session": "1"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+</code></pre>
+</div>
+
+At this time, the `/loginAs` endpoint does not run the `beforeLogin` or
+`afterLogin` hooks that would be invoked when calling the `/login` endpoint.
```
mtrezza: Let's remove "at time time", because it isn't clear what this means. If _never_ invokes the triggers as it's currently designed. We don't anticipate future features in the docs, since they may never occur (in that form). Once this behavior changes, we'll update the docs.

Can you change `hooks` to `triggers`, as they should be officially called in the docs. We could also link to the [trigger section](https://docs.parseplatform.org/cloudcode/guide/#session-triggers) for convenience.

We can omit to describe what other endpoints do, since they are not part of this chapter and we essentially duplicate information. We can write this more concise as:

> Calling this endpoint does not invoke the [session triggers](https://docs.parseplatform.org/cloudcode/guide/#session-triggers) such as `beforeLogin` and `afterLogin`.

_includes/rest/users.md
```diff
@@ -613,3 +611,53 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+
+Since the master key must always be supplied with any request to `/loginAs`,
+this action will always succeed if the supplied user id exists in the database.
```
mtrezza: I think it is already clear from the paragraph above that the master key is required. I would simply say that  the session is created regardless of account lockout.

mtrezza (changes requested):

_includes/rest/users.md
```diff
@@ -613,3 +611,47 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+## Impersonating a user
+
+An application may have use cases where an operator or automated system needs to
+take action on a user's behalf, under the user's authority, without access to
+the user's credentials. The Parse REST API supports these use cases with the
+`/loginAs` endpoint. This endpoint takes a `userId` parameter, and must be
+called using the master key. It will create a session for the given user ID, and
+will return the same response format as the `/login` endpoint. The new session
+will have a `createdWith` `action` value of `login`, and an `authProvider` value of
+`masterkey`.
```
mtrezza: I suggest the following edit to make this more concise and remove some ambiguity:

> An application may allow a user to take action on behalf of another user, without having access to the other user's login credentials. The Parse REST API provides the `/loginAs` endpoint which takes a `userId` parameter and creates a session for the given user. The created session has the property `createdWith: { action: 'login', authProvider: 'masterkey' }`. The endpoint must be called using the master key and returns the same response format as the `/login` endpoint.

Do you think it is clear that `userId` refers to a Parse User's `objectId`? How do the docs refer to this usually?

_includes/rest/users.md
```diff
@@ -613,3 +611,47 @@ For example, if you want the user with id `"3KmCvT7Zsb"` to have read and write
...
+       "X-Parse-Master-Key": "${MASTER_KEY}",
+       "X-Parse-Revocable-Session": "1"
+     })
+result = json.loads(connection.getresponse().read())
+print result
+</code></pre>
+</div>
+
+The `/loginAs` endpoint does not run the `beforeLogin` or `afterLogin` hooks
+that would be invoked when calling the `/login` endpoint.
```
mtrezza: I think we can omit to describe what other endpoints do, since they are not part of this chapter and we essentially duplicate information. We can add a link to the triggers and write this more concise as:

> Calling this endpoint does not invoke the [session triggers](https://docs.parseplatform.org/cloudcode/guide/#session-triggers) such as beforeLogin and afterLogin.

mtrezza (changes requested):

_includes/rest/users.md
GormanFletcher: The docs references to user IDs that I see are either assigning a variable `user_id = ...` in a code snippet, references to an `objectId` sitting beside the JSON printout for a user record, or references to the `id` property on a `User` object.

I think it's reasonable to clarify here that the `userId` parameter is the `objectId` on a `User` record.

mtrezza (approved):
LGTM, thanks for this PR!

TomWFox (approved):
LGMT 👍

dblythy:
Is this ready for release @mtrezza

mtrezza (approved):

PR #845, opened 2021-09-02
feat: add best practice page

mtrezza:
Add best practice page.

This PR is just to add a page so others can extend it. It does not claim to be complete by any means.

There is an extensive list of optimizations in the forum, but a forum post is likely to become harder to find and partially obsolete over time, so these would be easy candidates to be added to this page:
https://community.parseplatform.org/t/share-your-performance-tips/1572/


mtrezza:
@dblythy Anything you want to add?

dblythy:
This looks good for a start! Perhaps in time this could link to internal / external blogs too.

mtrezza:
Yes, I think we can link to external articles if they are cohesively addressing a certain topic. If they are just another list that is actually a substitute for this list (like the forum post I mentioned) I would not link, because it discourages people from extending this list, and the issue with forum and blog posts is that they are not editable by the community, so at some point they will become partly obsolete.

PR #847, opened 2021-09-22
Recommend `axios` instead of `request`

dblythy:
Request is depreciated and has been for 2 years. We shouldn't be recommending it as a third party module.


mtrezza:
Is this ready for review?

dblythy:
Yep, I don't have permission to request reviews though

mtrezza (approved):

PR #850, opened 2021-10-01
Update link in deleting the files

codewithdev:
Closing #838 

#### What thing being changed:

The documentation for "[Deleting the files](https://docs.parseplatform.org/rest/guide/#deleting-files)" contains invalid URL. This can be fixed by changing the URL from https://YOUR.PARSE-SERVER.HERE/parse/files/...profile.png to https://YOUR.PARSE-SERVER.HERE/parse/files/profile.png

### DIFF
--- a/_includes/rest/files.md
+++ b/_includes/rest/files.md
@@ -121,7 +121,7 @@ Users holding the master key are allowed to delete files using the REST API. To 
 curl -X DELETE \
   -H "X-Parse-Application-Id: <span class="custom-parse-server-appid">${APPLICATION_ID}</span>" \
   -H "X-Parse-Master-Key: <span class="custom-parse-server-masterkey">${MASTER_KEY}</span>" \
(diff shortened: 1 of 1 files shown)


mtrezza:
Thanks for this PR! Could you incorporate an explanation, as described in the issue?

> The documentation can be improved quickly by adding a clear example to explain how to format the URL.
>
>For example :
>
>Although the file's URL is 'http:////YOUR.PARSE-SERVER.HERE/parse/files/appId/my_picture.png', you have to remove the '/appId' section from the URL to access the API endpoind.
>
>So the URL must be : 'http:////YOUR.PARSE-SERVER.HERE/parse/files/my_picture.png'
>
>Otherwise it will not work.

codewithdev:
Definitely!
The images are located in the directory of parse which is *_/files_* and whenever send DELETE request to the given URL( https://YOUR.PARSE-SERVER.HERE/parse/files/...profile.png) which is an invalid format of the URL. Although, "appId" is required to retrieve the images which needs to be deleted.

mtrezza:
Yes, I think a short explanation that appId is required for retrieving but not deleting would be good in the docs. As I read the original issue, that missing info created the difficulty.

PR #881, opened 2022-06-29
docs: php hint for using master key for requests

cyb3rko:
While implementing the Parse PHP SDK an error occured that the requests are unauthorized even though we have specified the master key.

After digging into the PHP source code we found the optional parameter `useMasterKey`. But it wasn't documented anywhere so I added a note myself to the docs.

Is this okay or is there something to improve?


dblythy:
Can you add some detail around how masterKey should not be stored in code and used on trusted clients only

cyb3rko:
Sorry, I can not.
I do not use the PHP SDK by myself.

cyb3rko:
@dblythy What's the alternative to using a master key in PHP?
Using client keys?

PR #888, opened 2022-07-26
Create docs for Facebook Auth config usage

dblythy:
Currently there are no docs for how to configure Parse Server for facebook.


_includes/parse-server/third-party-auth.md
```diff
@@ -59,6 +59,18 @@ Note, most of them don't require a server configuration so you can use them dire
   }
 }
 ```
+
+The options passed to Parse Server:
+```js
+{
+  auth: {
+    facebook: {
+      appIds: "", // can be a string or an array of strings
```
mtrezza: `appIds: ["id1", "id2"],`

Could you also describe what this parameter does? And whether it is optional or not (I think it is optional)

mtrezza (changes requested):

_includes/parse-server/third-party-auth.md
mtrezza:
```suggestion
      appIds: ['appId1', 'appId2'], // If set, the app ID is used to validate the authentication token provided by the client when authenticating.
```

mtrezza (approved):

_includes/parse-server/third-party-auth.md
mtrezza: done

PR #896, opened 2022-08-07
docs: REST API python examples update

### LINKED ISSUE #894: 📙 Refurbishment of Python code examples
__Link to section:__
I don't know how many doc entries contain Python code but in the REST API guide are most of them:
https://docs.parseplatform.org/rest/guide

__What is the issue?__
The shown Python 2 code is outdated and gives instant errors if you try to run it in Python 3 (which is out since 2008, so definitely the today's standard):

image

To just make it work again it would be

```python
import json,http.client   # new import
connection = http.client.HTTPSConnection('YOUR.PARSE-SERVER.HERE', 443)   # using new import
connection.connect()
connection.request('GET', '/parse/classes/GameScore', '', {
    "X-Parse-Application-Id": "${APPLICATION_ID}",
    "X-Parse-REST-API-Key": "${REST_API_KEY}"
})
result = json.loads(connection.getresponse().read())
print(result)   # changed print syntax
```

But according to the PEP 8 coding guidelines the [imports should be on seperate lines](https://peps.python.org/pep-0008/#imports) and there should be [2 blank lines after the import section](https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention/47448739):

```python
import http.client
import json


...
```
---

Do we want to renew all the Python examples step by step to have modern examples you can directly use without errors and Googling what to change to make everything work again???

I'm interested in doing this if you say we want this to be done.

cyb3rko:
Updated the Python examples (Closes #894).
Changes I've made:

- replace Python 2 library `httplib` with newer `http.client` library
- fix invalid print commands for Python 3 (`print result` -> `print(result)`
- adapt `import` syntax to coding conventions
- adapt `urllib.parse.urlencode` parameters to coding conventions

Optional step left:
- fixing wrong indentations

Do we want to fix it or was there any reason why it was implemented like that?

image

### DIFF
--- a/_includes/rest/analytics.md
+++ b/_includes/rest/analytics.md
@@ -34,8 +34,11 @@ curl -X POST \
   <span class="custom-parse-server-protocol">https</span>://<span class="custom-parse-server-url">YOUR.PARSE-SERVER.HERE</span><span class="custom-parse-server-mount">/parse/</span>events/AppOpened
 </code></pre>
 <pre><code class="python">
-import json,httplib
-connection = httplib.HTTPSConnection('<span class="custom-parse-server-url">YOUR.PARSE-SERVER.HERE</span>', 443)
+import http.client
+import json
+
+
+connection = http.client.HTTPSConnection('<span class="custom-parse-server-url">YOUR.PARSE-SERVER.HERE</span>', 443)
 connection.connect()
 connection.request('POST', '<span class="custom-parse-server-mount">/parse/</span>events/AppOpened', json.dumps({
      }), {
@@ -44,7 +47,7 @@ connection.request('POST', '<span class="custom-parse-server-mount">/parse/</spa
        "Content-Type": "application/json"
      })
 result = json.loads(connection.getresponse().read())
-print result
+print(result)
 </code></pre>
 </div>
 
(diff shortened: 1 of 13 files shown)


mtrezza:
Looks to me as if the indentation was by mistake. There may be some IDEs that apply special indentation rules by default, one example is Android Studio. But indentation is generally a custom style decision, so I'd fix the indentation for better readability and consistency.

cyb3rko:
Yes, most modern IDEs support auto-formatting.

Then I will fix this as well. The Python conventions say that one indentation level equals 4 spaces:
https://peps.python.org/pep-0008/#indentation

cyb3rko:
Alright, fixed indentation and Json formatting.

cyb3rko:
But I realised a few of those examples have a fix domain and don't support that input at the top of the page where you enter your own server connection information.

Is that intended or should we fix this as well?
Example: https://docs.parseplatform.org/rest/guide/#localizing-push

mtrezza:
I think this is a well-intended feature, but I don't think we should provide it for security reasons.

This feature asks developers to enter their Parse Server URL, master key and client keys on a webform and submit it. Asking that of a developer goes against establishing a good security practice and facilitates phishing. IMO we should never ask a developer to enter this information anywhere, but in fact create awareness about the sensitivity of that data and remind to never share it with anyone outside their project.

The only way such a feature may make sense was if the docs were made part of a Parse Dashboard backend where the user logged into the dashboard already has access to that information and it is merely displayed from the backend data.

I'm for removing this feature from the docs and just use common placeholders throughout the code. I'd do that in a separate PR though.

Opened https://github.com/parse-community/docs/issues/898 to track this.

cyb3rko:
Alright, then it's ready to merge.

mtrezza (approved):
Looks good!

PR #904, opened 2022-10-11
docs: fix incorrect query constraint

k4vrin:
Queries on String Values -> Full Text Search.
In the code example, the value that has been passed is "Big Daddy's" but the description is about "bbq" word in the "name" String key


mtrezza (approved):
Looks good! This is somewhat confusing because it doesn't show a data table in the code example. That could be helpful.

PR #909, opened 2022-12-22
docs: change naming for swift function 

vazarkevych:
- change the naming swift function from 'setDefaultACL' to 'setDefault'
- change the naming swift function from 'handlePush' to 'handle'


mtrezza:
Related to https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1683

PR #910, opened 2022-12-22
feat: use npm lockfile v2

Moumouls:


Moumouls:
Here the lockfile V2 PR @mtrezza

PR #918, opened 2023-01-31
docs: updated readme to use includes folder as starting directory

RahulLanjewar93:
The current doesn't mention the starting directory to be _includes.
Have mentioned that the starting directory is _includes


mtrezza (approved):
Looks good!

PR #919, opened 2023-01-31
docs: update incorrect batch size reference

RahulLanjewar93:
In current parse-server, I dont think there is a cap for how many objects can be updated/saved/destroyed in one go.
So maybe we should update this in docs.

### DIFF
--- a/_includes/rest/objects.md
+++ b/_includes/rest/objects.md
@@ -501,7 +501,7 @@ print(result)
 
 ## Batch Operations
 
-To reduce the amount of time spent on network round trips, you can create, update, or delete using the batch endpoint.
+To reduce the amount of time spent on network round trips, you can create, update, or delete using the batch endpoint. The batch size can be customized, the default batch size is 20.
 
 Each command in a batch has `method`, `path`, and `body` parameters that specify the HTTP command that would normally be used for that command. The commands are run in the order they are given. For example, to create a couple of `GameScore` objects:
 
--- a/_includes/rest/objects.md
+++ b/_includes/rest/objects.md
@@ -501,7 +501,7 @@ print(result)
 
 ## Batch Operations
 
-To reduce the amount of time spent on network round trips, you can create, update, or delete up to 50 objects in one call, using the batch endpoint.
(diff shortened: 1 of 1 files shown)


mtrezza (commented):
- What is the current batch size limit?
- Is this the only place where this limit is mentioned in the docs?

RahulLanjewar93:
- The current batch size is 20.   `REQUEST_BATCH_SIZE: 20` https://github.com/parse-community/Parse-SDK-JS/blob/10a2e5b02db62bdb5edd996ff05198cd96d127f5/src/CoreManager.js#L187
- I am not sure , but I think batch endpoints are not mentioned in other platforms

RahulLanjewar93:
Also, there is [this](https://github.com/parse-community/Parse-SDK-JS/pull/1090) PR on how to use `transactions` on client sdks while using the batch endpoint.
It's not merged yet and somewhere they did mention removing the batch size and putting everything in a single batch because it's causing issues.

mtrezza:
If the current batch size is 20, not 50, should that be mentioned in the docs instead of removing the note altogether?

RahulLanjewar93:
@mtrezza the batch size is customizable in the JS SDK. The default is set to 20.

_includes/rest/objects.md
```diff
@@ -501,7 +501,7 @@ print(result)
 
 ## Batch Operations
 
-To reduce the amount of time spent on network round trips, you can create, update, or delete up to 50 objects in one call, using the batch endpoint.
+To reduce the amount of time spent on network round trips, you can create, update, or delete using the batch endpoint.
```
mtrezza: Thank for looking this up, should we add this?

```suggestion
To reduce the amount of time spent on network round trips, you can create, update, or delete using the batch endpoint. The batch size can be customized, the default batch size is 20.
```

RahulLanjewar93: Yes, that makes sense. I'll add that.

mtrezza: Oh you could just have commited the suggestion with the button above :-)

mtrezza (approved):
Looks good!

PR #923, opened 2023-04-14
feat: change startup method

### LINKED ISSUE #922: Update to PS6 📙
__Link to section:__
Please provide a specific link to the section in question. In the docs, you can click on a heading and then copy the relative link from the address bar.

Docs still use older PS5 syntax

__What is the issue?__
Please provide a detailed explanation of the problem, quoting parts of the docs or using screenshots if appropriate.

dblythy:
Closes #922


mtrezza (approved):
Looks good!

Issue #811, opened 2021-01-05
💡 Expand “Best practices” guide


dblythy:
There are a number of issues that have been raised regarding firewall or other stack related issues that cannot be solved by the Parse Server package. 

I think to help our community, I propose a “best practices / recommended configuration (or even FAQ)” section. From the top of my head, I’d imagine:

-[Rate Limiting](https://github.com/parse-community/parse-server/issues/1505) - Parse wide / specific endpoints e.g /requestPasswordReset
-[Improving Performance](https://community.parseplatform.org/t/what-someone-should-do-to-increase-performance-of-parse-server/973/10)
-[Unit testing cloud code](https://github.com/parse-community/parse-server/issues/2488)

It would be good to try to get some shared knowledge going from how the core team’s configure their own servers, especially as I’d imagine most of our developers aren’t overly experienced with application stacks.

mtrezza:
There have been threads on GitHub and/or Parse Forum regarding this. It could be helpful if you link them here to consolidate the discussion.

In addition, there have been troubleshooting threads that you could draw material from for such a best practice document.

For example:
- https://github.com/parse-community/parse-server/issues/6778

dblythy:
Of course, I've edited the original post to include some links. If you have any suggestions for topics that could be included in the section, please feel free to post them too.

mtrezza:
Parse Server docs do have a [Best Practice](https://docs.parseplatform.org/parse-server/guide/#best-practice) page.

PR #924, opened 2023-04-14
feat: expand best practices

dblythy:
Related: #811


_includes/parse-server/best-practice.md
```diff
@@ -10,6 +10,57 @@ Protect all Parse Server endpoints using a Firewall to mitigate the risk of mali
 - Use rate-limiting rules for public endpoints, for example limit the number of requests per IP address or per user.
 - Use very restrictive rules for private endpoints; for example limit access to Parse Dashboard to your personal network.
 
+### Security-first mindset
```
mtrezza:
```suggestion
### Security-First Mindset
```

_includes/parse-server/best-practice.md
```diff
@@ -10,6 +10,57 @@ Protect all Parse Server endpoints using a Firewall to mitigate the risk of mali
...
+
+When developing Parse Server, assume that any cloud function or custom trigger is going to be abused by a malicous client. It's best to assume that the client code cannot be trusted. Even if you have written the client code, it's very simple for an attacker to replicate a request to your Parse Server and modify arguments.
+
+The following cloud code is **not recommended** and is an example of poor security practise:
```
mtrezza:
```suggestion
When developing for Parse Server, consider any Cloud Function or Cloud Trigger as a potential target for a malicious attack. In general, it's best to always assume that the sender of a request cannot be trusted. Even if you have control over the client's code when shipping it, it's simple for an attacker to replicate a request to Parse Server and modify arguments, or reverse-engineer a client app and extract hard coded keys to send requests with unexpected arguments.

The following cloud code is **not recommended** and is an example of poor security practise:
```

_includes/parse-server/best-practice.md
```diff
@@ -10,6 +10,57 @@ Protect all Parse Server endpoints using a Firewall to mitigate the risk of mali
...
+    const user = await Parse.Cloud.run('updateEmail', {id});
+    console.log(`Here is a full user object: `, user.toJSON());
+    return;
+  } catch (e) {
+    console.log(`Not a valid id: ${id}`);
+  }
+}
+```
+
+Regardless of how secure your Parse Server is with ACLs and CLPs, a few lines of insecure cloud functions can undo it.
```
mtrezza:
```suggestion
Regardless of how secure your Parse Server is with ACLs and CLPs, a few lines of insecure Cloud Functions can undo it.
```

_includes/parse-server/best-practice.md
```diff
@@ -10,6 +10,57 @@ Protect all Parse Server endpoints using a Firewall to mitigate the risk of mali
 - Use rate-limiting rules for public endpoints, for example limit the number of requests per IP address or per user.
 - Use very restrictive rules for private endpoints; for example limit access to Parse Dashboard to your personal network.
 
+### Security-First Mindset
+
+When developing for Parse Server, consider any Cloud Function or Cloud Trigger as a potential target for a malicious attack. In general, it's best to always assume that the sender of a request cannot be trusted. Even if you have control over the client's code when shipping it, it's simple for an attacker to replicate a request to Parse Server and modify arguments, or reverse-engineer a client app and extract hard coded keys to send requests with unexpected arguments.
```
mtrezza, line 15: This looks like something I've written some time ago, if I'm not mistaken. Where did you copy that from? Are we duplicating content with this?

mtrezza:
Should we merge this?

dblythy:
It's ready for merge if you are ok!

_includes/parse-server/best-practice.md
```diff
@@ -10,6 +10,57 @@ Protect all Parse Server endpoints using a Firewall to mitigate the risk of mali
...
+```
+
+Regardless of how secure your Parse Server is with ACLs and CLPs, a few lines of insecure Cloud Functions can undo it.
+
+A secure solution would be to ensure that users can only update their email, such as:
+
+```js
+Parse.Cloud.define('updateEmail', async (req) => {
+  req.user.set('email', req.params.email);
+  await req.user.save(null, { useMasterKey: true });
```
mtrezza, line 56: Wouldn't it be best practice to avoid any operation using `useMasterKey: true` but instead give only the user write permission in the object ACL? This would mitigate the risk that through a developer mistake in code someone could execute operations because they are using the master key; instead any permission would be bound to the user's session token. Using `useMasterKey: true` should only be necessary if the object is locked even for the user with read/write limited to master key.

PR #930, opened 2023-05-25
fix: flutter svg icon and flutter document and php api path

mbfakourii:
The Flutter icon is not shown. Also, there were small problems in the Flutter document
Also, there was a problem with the php api path

https://docs.parseplatform.org/


mtrezza (approved):
Looks good

Issue #48, opened 2015-05-18
Restore Tasks Section in Unity Guide


hramos:
The Task section of the Unity SDK Docs is missing:
https://parse.com/docs/unity/guide#tasks

BenWoodford:
Similarly all the examples that use tasks are using the .NET snippets, aka using await and async which we don't have the pleasure of using in Unity.

facebook-github-bot-3:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

hallucinogen:
@BenWoodford can you make another issue and assign it to me?

IIRC, Task section was purposely deleted from our Unity SDK Docs since we (and Unity) don't officially support `System.Threading.Tasks`. I'll revisit this soon.

PR #931, opened 2023-05-25
feat: Update Dart, Flutter API links and improve document structure

mbfakourii:
There was a problem with the flutter `API Reference`

https://docs.parseplatform.org/flutter/guide/


flutter.md
```diff
@@ -5,7 +5,7 @@ layout: guide
 platform: flutter
 language: dart
 display_platform: Flutter
-api_reference: https://parse-community.github.io/Parse-SDK-Flutter/api
+api_reference: https://pub.dev/documentation/parse_server_sdk_flutter/latest/
```
mtrezza: I think the API reference would be: https://pub.dev/documentation/parse_server_sdk_flutter/latest/flutter_parse_sdk_flutter/flutter_parse_sdk_flutter-library.html

But actually I think we should generate our own docs and publish them like we do with our other products, instead of redirecting to a 3rd party website.

mtrezza (changes requested):

flutter.md
mbfakourii: Yes, I think the same

But I think it is better to publish in this way until the api is created

mbfakourii: I made the changes

mtrezza:
I've added docs publishing to the Flutter SDK; the paths are:
- http://parseplatform.org/Parse-SDK-Flutter/dart/
- http://parseplatform.org/Parse-SDK-Flutter/flutter/

These are the roots of the generated docs, but since we don't need to land on the README, I'd suggest to link directly to the API docs:

- http://parseplatform.org/Parse-SDK-Flutter/dart/flutter_parse_sdk/flutter_parse_sdk-library.html
- http://parseplatform.org/Parse-SDK-Flutter/flutter/flutter_parse_sdk_flutter/flutter_parse_sdk_flutter-library.html

Also, there's still a minor issue with the product logo not resizing in the README to HTML conversion, so another reason to link directly to the API docs:
image

mbfakourii:
> I've added docs publishing to the Flutter SDK; the paths are:
> 
> * http://parseplatform.org/Parse-SDK-Flutter/dart/
> * http://parseplatform.org/Parse-SDK-Flutter/flutter/
> 
> These are the roots of the generated docs, but since we don't need to land on the README, I'd suggest to link directly to the API docs:
> 
> * http://parseplatform.org/Parse-SDK-Flutter/dart/flutter_parse_sdk/flutter_parse_sdk-library.html
> * http://parseplatform.org/Parse-SDK-Flutter/flutter/flutter_parse_sdk_flutter/flutter_parse_sdk_flutter-library.html
> 
> Also, there's still a minor issue with the product logo not resizing in the README to HTML conversion, so another reason to link directly to the API docs: image

Regarding the logo, we will solve the problem in Repastory Flutter. In this issue, we can use your links.

Added `api.html` page to solve the problem

image

mtrezza:
> Added api.html page to solve the problem

What is the purpose of this extra page? I don't think a page with just 2 links is needed as our current structure throughout the website is to look at them as 2 separate SDKs. In addition, if the Flutter SDK docs include the Dart SDK classes as you mentioned previously, this page makes even less sense I think?

mbfakourii:
> > Added api.html page to solve the problem
> 
> What is the purpose of this extra page? I don't think a page with just 2 links is needed as our current structure throughout the website is to look at them as 2 separate SDKs. In addition, if the Flutter SDK docs include the Dart SDK classes as you mentioned previously, this page makes even less sense I think?

It makes sense from this point of view but what do you suggest to link only the Flutter version? Or should we add two versions of Dart to the document?

mbfakourii:
I don't think two documents are good because they are 90% alike!

I have a suggestion, for example, two documents should be added in the image below, but their guide should be for one page

241227301-a858386c-d972-4238-ad3c-c67dde49ffaf

mbfakourii:
See [this page](https://github.com/parse-community/docs/tree/f2b7788d279de705b60c4263dab74486c1aa5d65/_includes/common)

The idea of the `common` folder can be reasonable, but I think it should have another name . for example `dartFlutterCommon`

What do you think?

mtrezza:
What are these 10% difference? Is that within 1 section or scattered across the docs?

I should note that we are thinking about combining the docs into a single document. Then in every section there will be tabs to switch between languages / SDKs for the code examples. There is too much duplicate content across the SDKs. But that is a separate, more complex effort. I just say this because it may make the current solution obsolete, so that we don't over-complicate things.

mbfakourii:
> What are these 10% difference? Is that within 1 section or scattered across the docs?
> 
> I should note that we are thinking about combining the docs into a single document. Then in every section there will be tabs to switch between languages / SDKs for the code examples. There is too much duplicate content across the SDKs. But that is a separate, more complex effort. I just say this because it may make the current solution obsolete, so that we don't over-complicate things.

I said about 90 percent. The difference that may be in the documents is the path of the library that is in the [getting-started.md](https://github.com/parse-community/docs/blob/f2b7788d279de705b60c4263dab74486c1aa5d65/_includes/flutter/getting-started.md)

To be precise, maybe 99% are similar!

If we want to put everything in single document, I think the best way is to use `api.html`

mtrezza:
You say 99% are similar, so there's only 1% difference, that is content that is in the Flutter docs but not in the Dart docs? Could you point out what this 1% is specifically?

It may be reasonable to combine the documentation for both into one and not distinguish between the two SDKs on a high level.

mbfakourii:
> You say 99% are similar, so there's only 1% difference, that is content that is in the Flutter docs but not in the Dart docs? Could you point out what this 1% is specifically?
> 
> It may be reasonable to combine the documentation for both into one and not distinguish between the two SDKs on a high level.

This 1% is related to Dart and Flutter package url which is at the beginning of the `getting-started.md` file

mbfakourii:
> You say 99% are similar, so there's only 1% difference, that is content that is in the Flutter docs but not in the Dart docs? Could you point out what this 1% is specifically?
> 
> It may be reasonable to combine the documentation for both into one and not distinguish between the two SDKs on a high level.

This 1% is related to Dart and Flutter library path which is at the beginning of the getting-started.md file

mtrezza:
The package URL is the only docs difference between the two SDKs? I understand that the Flutter SDK is built on the Dart SDK, but I assume there are other things that the Flutter SDK adds on top of the Dart SDK that would be documented?

mbfakourii:
> The package URL is the only docs difference between the two SDKs? I understand that the Flutter SDK is built on the Dart SDK, but I assume there are other things that the Flutter SDK adds on top of the Dart SDK that would be documented?

Yes, I noticed this when I was merging the documents . Maybe there are things in Flutter that are not documented !

mtrezza:
Are you referring to the generated docs (dart doc), or the guide?

mbfakourii:
> Are you referring to the generated docs (dart doc), or the guide?

I mean the [dart](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/dart/README.md) and [flutter](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/flutter/README.md) guides that are now available in the Flutter repository

mtrezza:
Maybe let's not over-complicate this, just like we decided to keep the Parse Dart SDK separate from the Parse Flutter SDK, we have two separate blocks for Dart / Flutter:

image

All I would do in this PR is:
- add the correct API links (currently pointing to pub.dev)
- add the correct Flutter guide link (currently pointing to the README)

There is no Parse Dart SDK guide so far, only a Parse Flutter SDK guide. Since you say they are 99% identical, we can either change the dart guide link to point to the flutter guide, or keep it pointing to the README.

I'd remove all other changes in this PR, such as this additional page:
image

I don't see that page adding much value, at first we distinguish between the 2 SDKs in the blocks (1st screenshot), then we have the API page (2nd screenshot), and when someone clicks in that page on "Flutter API Ref" they get to an API documentation where Dart and Flutter is mixed again. Confusing.

We can deal with the fact that the docs are 99% identical in a separate issue and discuss it before opening a PR. One possible solution could be to use a `common` directory where the 99% identical files live and then compose the two separate docs based on that, to avoid content duplication. Or we can complete merge the docs for dart and flutter and  remove that distinction throughout the docs. I'd discuss this in a separate issue.

mbfakourii:
I added the `dart`

image

and deleted the `api.html` page

I noticed that the `push-notifications.md` page is not supported in Dart, so I did not add it to the Dart pages

API Reference paths were also set correctly

mtrezza:
The changes need to be on the front page as well. The _docs.parseplatform.org_ page seems outdated as I mentioned previously and I'm unsure whether it's used anywhere at all. It's a duplication of what's on the front page, but an outdated one. Fixing this is a separate docs issue.

The front page on the other hand already has a Dart section, but the links there need to be updated, for Dart and Flutter.

image

mbfakourii:
@mtrezza 
I think the front page is in [another repository](https://github.com/parse-community/parse-community.github.io)

mbfakourii:
fixed
please check [#48](https://github.com/parse-community/parse-community.github.io/pull/48)

mtrezza (commented):
This looks good, can we merge?

mtrezza (approved):

mbfakourii:
> This looks good, can we merge? 

Yes

PR #934, opened 2023-06-24
docs: document TOTP adapter

dblythy:


_includes/parse-server/third-party-auth.md
```diff
@@ -405,3 +407,168 @@ For more information about custom auth please see the examples:
...
+});
+```
+
+Now, MFA will be enabled for the user. You can access recovery keys by:
+
+```js
+const recovery = user.get("authDataResponse");
+```
+
+It's also recommended to clear the authData from the client side:
```
mtrezza: Could you add why this is recommended?

_includes/parse-server/third-party-auth.md
```diff
@@ -405,3 +407,168 @@ For more information about custom auth please see the examples:
...
+    });
+  } catch (e) {
+    // display error
+  }
+}
+```
+
+### SMS One-Time passwords
+
+It is recommended to use TOTP MFA over SMS OTPs as SMS OTPs can be intercepted via [SS7](https://www.theguardian.com/technology/2016/apr/19/ss7-hack-explained-mobile-phone-vulnerability-snooping-texts-calls) or [sim swap](https://us.norton.com/blog/mobile/sim-swap-fraud) attacks.
```
mtrezza: We don't link to 3rd party websites of such nature in the docs. Could you extract the technical term for this interception technique (if there is any) so people can search online for details if they are interested? Otherwise the note that it can be intercepted is detailed enough.

_includes/parse-server/third-party-auth.md
```diff
@@ -405,3 +407,168 @@ For more information about custom auth please see the examples:
...
+await user.save({ authData: { mfa: { mobile: "+11111111111", token: code } } });
+```
+
+Now, SMS MFA will be enabled for the user. You can access recovery keys by accessing:
+
+```js
+const recovery = user.get("authDataResponse");
+```
+
+It's also recommended to clear the authData from the client side:
```
mtrezza: Again, could you add why?

_includes/parse-server/third-party-auth.md
```diff
@@ -405,3 +407,168 @@ For more information about custom auth please see the examples:
...
+      period: 30,
+    },
+  },
+}
+```
+
+To enable SMS MFA for a user, first set the users' mobile number.
+
+```js
+await user.save({ authData: { mfa: { mobile: "+11111111111" } } });
```
mtrezza: It seems there is no official standard by the ITU for example phone numbers, like example.com by the IANA. I'd suggest to stick with the most common example phone number which is `+15555555555`. Twilio [docs](https://www.twilio.com/docs/voice/trusted-calling-with-shakenstir#use-a-calltoken-when-forwarding-a-call) also use this primarily, while in examples with multiple phone numbers they use the digits 2 and 8.

> A caller (+12222222222) dials your Twilio phone number (+15555555555) and you need to forward the call to a new phone number (+18888888888).

mtrezza (changes requested):

mtrezza (approved):
Looks good, is this ready for merge?

PR #937, opened 2023-09-28
Add ios docs for Swift Package Manager

dplewis:
Cocoapods and Carthage are no longer supported. Add documentation for installing via SPM.

See https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1740


mtrezza (approved):
Looks good

PR #940, opened 2023-10-20
docs: Fix google auth for Parse

luca-corteccioni:
In ParseUser.loginWith('google', google( ... , ... , ...) ), the variables of the google() function are placed in incorrect order, which causes error "status code 101, auth is invalid for this user" from parse.

### DIFF
--- a/_includes/dart/users.md
+++ b/_includes/dart/users.md
@@ -94,8 +94,8 @@ class OAuthLogin {
     GoogleSignInAuthentication authentication = await account.authentication;
     await ParseUser.loginWith(
         'google',
-        google(_googleSignIn.currentUser.id, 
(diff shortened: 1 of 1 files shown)


mtrezza:
Thanks for the PR, do you have a docs reference to verify the order?

luca-corteccioni:
Yes is in

luca-corteccioni:
You can see that in file parse_login_helper.dart of flutter_parse_sdk there is a function that generate the authData to pass:
```
Map<String, dynamic> google(String token, String id, String idToken) {
  return <String, dynamic>{
    'access_token': token,
    'id': id,
    'id_token': idToken
  };
}
```

mtrezza (approved):
Looks good!

PR #942, opened 2024-01-10
Update usage.md

jaysonng:
For await to work, it needs to be inside an async function. or else an error will occur 

```SyntaxError: await is only valid in async functions and the top level bodies of modules```


mtrezza:
That's only true for old Node.js versions

jaysonng:
interesting. I'm running node 18. 

is that for node 20? and what can we do you think to add to these docs for subtle differences in node version code?

Issue #947, opened 2024-05-07
📕Provide information on using Facebook's Limited Token


chriscborg:
__References:__
[Changes facebook auth host to resolve JWT validation issue](https://github.com/parse-community/parse-server/pull/9122)

__Proposed solution:__
This current documentation does not specify how a user can use Facebook's JWT token provided with Limited Login to authenticate.

PR #948, opened 2024-05-07
docs: Update docs to include authentication with Facebook Limited Login

chriscborg:
As suggested in issue #947, this updates documentation to include information on what data to pass to the SDK in order to authenticate with Facebook Limited Login's JWT token. Related to issue: https://github.com/parse-community/parse-server/issues/9117


_includes/parse-server/third-party-auth.md
```diff
@@ -60,6 +60,17 @@ Note, most of them don't require a server configuration so you can use them dire
 }
 ```
 
+With Facebook's iOS SDK 17, you are required to implement Facebook Limited Login. If the user does not allow tracking through Apple's App Tracking Transparency, then the Facebook returns a JWT token instead of an access token. Therefore, in your app you would need to check if tracking is allowed or not and pass the relevant tokens. If your app does not receive an access token, then you will need to pass the below instead. *Available on parse-server v6.5.6 and above and v7.0.1 and above.*
```
mtrezza:
```suggestion
With Facebook's iOS SDK 17, you are required to implement Facebook Limited Login. If the user does not allow tracking through Apple's App Tracking Transparency, then the Facebook returns a JWT token instead of an access token. Therefore, in your app you would need to check if tracking is allowed or not and pass the relevant tokens. If your app does not receive an access token, then you will need to pass the below instead. *Available on Parse Server >= 6.5.6 < 7 and >=7.0.1.*
```

mtrezza (changes requested):
server version fix

mtrezza (approved):
Looks good; just waiting for the Parse Server PRs to get merged, then we merge this.

